PR #411, opened 2024-07-04
Replace bs4 beautifulsoup4

fabaff:
`bs4` is a dummy package that points to `beautifulsoup4`. 

Most distributions don't ship `bs4`.


ocervell:
Thanks a lot ;)

PR #426, opened 2024-09-13
feat(`dnsx`): add IP output type on dnsx A record 

romisfrag:
added DNS A record that produce an IP output type in addition of the A record output type


secator/tasks/dnsx.py
```diff
@@ -47,6 +47,12 @@ def item_loader(self, line):
 				if isinstance(value, dict):
 					name = value['name']
 					extra_data = {k: v for k, v in value.items() if k != 'name'}
+				# Handling specific records that can produce some other output types
```
ocervell:
```suggestion

```

ocervell (approved):

PR #428, opened 2024-09-14
dnsx: added reverse dns lookup

romisfrag:
Added  reverse dns lookup for dnsx and remove the -ptr for the default command because after some test it seems useless (maybe I am wrong but the commit removes it so make sure to put it back if needed)


secator/utils.py
```diff
@@ -439,3 +439,26 @@ def print_version():
 	console.print(f'[bold gold3]Lib folder[/]: {LIB_FOLDER}')
 	if status == 'outdated':
 		console.print('[bold red]secator is outdated, run "secator update" to install the latest version.')
+
+
+def extract_root_domain_from_domain(domain):
```
ocervell: This definitely requires a unittest to check if it catches all the cases (weirdly formatted domains, multi-level subdomains, domains with special characters in the name, etc...)

ocervell:
I've updated the PR to do reverse lookup by default (-ptr always activated like before) and to explicitely yield output types (see https://github.com/freelabz/secator/pull/439 for context)

PR #430, opened 2024-09-14
httpx: added support for -tls_grab

romisfrag:
httpx: added support for -tls_grab which look into certificate subject name and alternative names for domain name


secator/tasks/httpx.py
```diff
@@ -68,6 +73,9 @@ class httpx(Http):
 	proxy_socks5 = True
 	proxy_http = True
 	profile = 'cpu'
+	# Overriding the output_type to add Subdomains
```
ocervell:
```suggestion
```

secator/utils.py
```diff
@@ -439,3 +439,46 @@ def print_version():
 	console.print(f'[bold gold3]Lib folder[/]: {LIB_FOLDER}')
 	if status == 'outdated':
 		console.print('[bold red]secator is outdated, run "secator update" to install the latest version.')
+
+
+def extract_root_domain_from_domain(domain):
```
ocervell: needs a unit test

secator/tasks/httpx.py
```diff
@@ -68,6 +73,9 @@ class httpx(Http):
...
+	output_types = [Url, Subdomain]
+	founded_domain_in_certificates = []
```
ocervell: this will cause issues (remote Celery run, task chunking, multiple tasks instantiations, etc...). Class attributes are not meant to be the task memory. Use the `on_init` hook and set `self.cert_domains = []` instead.

secator/tasks/httpx.py
```diff
@@ -83,17 +91,50 @@ def on_init(self):
...
+					if v[-2:] == 'ms':
+						response_time = response_time / 1000
+					item[k] = response_time
+				elif k == URL:
+					item[k] = sanitize_url(v)
+			item[URL] = item.get('final_url') or item[URL]
+		# Because item can also be a domain
+		except KeyError as e:
+			if not ('url' in e.args):
+				raise e
```
ocervell: I'm not sure this is needed; this is run on item pre-convert so you shouldn't have any secator output type at this point.

secator/tasks/httpx.py
```diff
@@ -83,17 +91,50 @@ def on_init(self):
...
 
+	def _create_subdomain_from_tls(self, domain_in_certificate):
+		if domain_in_certificate.startswith('*'):
+			domain_in_certificate = remove_first_subdomain(domain_in_certificate)
+		if not (domain_in_certificate in self.founded_domain_in_certificates):
+			self.founded_domain_in_certificates.append(domain_in_certificate)
+			return {
+				'host': domain_in_certificate,
+				'domain': extract_root_domain_from_domain(domain_in_certificate),
+				'_source': 'httpx'
```
ocervell:
```suggestion
```

secator/tasks/httpx.py
```diff
@@ -83,17 +91,50 @@ def on_init(self):
...
+		for k, v in item.items():
+			if k == 'tls':
+				for tls_k, tls_v in v.items():
+					if tls_k == 'subject_cn':
+						domain = self._create_subdomain_from_tls(tls_v)
+						yield domain
+					if tls_k == 'subject_an':
+						for alternative_name in tls_v:
+							domain = self._create_subdomain_from_tls(alternative_name)
+							yield domain
```
ocervell:
```suggestion
	def on_json_loaded(self, item):
	    """Extract domain from TLS certificate and yield a Subdomain if present."""
	    yield item
	    tls = item.get('tls', None)
	    if tls:
		    for key, value in tls.items():
			    if key== 'subject_cn':
				    domain = self._create_subdomain_from_tls(value)
				    yield Subdomain(**domain)
			    if key== 'subject_an':
				    for alternative_name in value:
					    domain = self._create_subdomain_from_tls(alternative_name)
					    yield Subdomain(**domain)
```

ocervell (changes requested):

secator/runners/_base.py
```diff
@@ -827,12 +827,18 @@ def _process_item(self, item: dict):
...
 
-		# Run item hooks
-		item = self.run_hooks('on_item_pre_convert', item)
-		if not item:
-			return None
-
 		# Convert output dict to another schema
+		if isinstance(item, OutputType):
+			pass
+		elif isinstance(item, dict):
```
ocervell:
```suggestion
		if isinstance(item, dict):
```

PR #459, opened 2024-10-20
feat(output_types): add Certificate output type

romisfrag:


romisfrag:
The lint tests don't pass for the Certificate_status class, I've tried an epsection but it does not work, if you have a working fix for that it would be great

romisfrag:
I don't get It, the lint and unit test are working locally, I don't have the same issues as reported by the github jobs.

PR #484, opened 2024-11-17
feat(task): testssl.sh integration

romisfrag:
Now still not working need to resolve this issue
image


romisfrag:
Also need to fix integration tests for testssl

ocervell:
@romisfrag I've changed the logic quite a bit, but the code is way cleaner now. Could you test and validate the behavior of the updated branch and let me know what you think ?

I've added a check for soon-to-expire certificates, and we are now yielding a medium vulnerability when the certificate is actually expired.

Some certificates have an 'UNKNOWN' status, so we might need to do a bit more work to check if we can retrieve a better status.

PR #604, opened 2025-04-29
fix: use distro.like() instead of distro.id()

### LINKED ISSUE #603: bug: distro not correctly detected on tools install
I have a Debian rolling distro named kaisen
When I try to install tools there is an error complaining about badly detected distro

```
[INF] Installing bbot
Traceback (most recent call last):
[REDACTED]
    distribution = get_distro_config()
  File "/home/user/.local/share/pipx/venvs/secator/lib/python3.13/site-packages/secator/installer.py", line 562, in get_distro_config
    manager = installer.split(' ')[0]
              ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'split'
```
https://github.com/freelabz/secator/blob/5e5f04cce7abb884d7bab68ea587d9df9fdb3129/secator/installer.py#L531-L536

According to distro, you should use `like` instead of `id`
https://pypi.org/project/distro/

Image

On my kaisen
```
>>> import platform
>>> system = platform.system()
>>> print(system)
Linux
>>> import distro
>>> distro.id()
'kaisen'
>>> distro.like()
'debian'
```

`like` is used to identify the parent family distro and is more reliable.
I've updated the code in my install to use `like` instead of `id` and it works well

psyray:
fix #603 

Use distro.like() instead of distro.id() to determine the Linux distribution. This handles cases where custom distro returns a not supported distro, ensuring the correct installer is used.


secator/installer.py
```diff
@@ -529,7 +529,7 @@ def get_distro_config():
 	distrib = system
 
 	if system == "Linux":
-		distrib = distro.id()
+		distrib = distro.like()
```
ocervell:
```suggestion
		distrib = distro.id() or distro.like()
```

secator/installer.py
```diff
@@ -559,6 +559,9 @@ def get_distro_config():
 		else:
 			installer = "scoop"  # Alternative package manager for Windows
 
+	if installer is None:
+		return None
+
```
ocervell:
```suggestion
```

PR #650, opened 2025-05-17
deploy: Kustomize deployment

0xlildoudou:
Hi,

Add a basic Kustomize deployment for Kubernetes

### DIFF
--- a/kustomize/README.md
+++ b/kustomize/README.md
@@ -1,0 +1,1 @@
+# Kustomize deployment for Secator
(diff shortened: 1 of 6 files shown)


ocervell:
Thanks !

Shouldn't we use the existing Helm chart for customize to tweak instead ? https://www.reddit.com/r/kubernetes/comments/11k4t06/kustomize_vs_helm_charts/

This will make it harder to maintain, if we have to make changes to both helm + kustomize deployments

PR #712, opened 2025-07-30
WIP feat(task): add enumdns

0xlildoudou:


ocervell:
Hi @0xlildoudou,

What does this tool brings that `dnsx` doesn't ? Is it worth integrating anyway ?

PR #713, opened 2025-09-16
fix(fping): remove bad options, add new ones

Moins-2:
## Remove options 

The option `--r` doesn't exist on `fping`. So i removed it. I also changed the `opt_prefix` to `-`, i don't really know on what was based the old task, but now it's coherent to `fping`.

## New options 

Added the options `-c`, `-n`, `-d` and `-s`
The `-c` / count option added a text to the output, making the parser not working anymore, in a more verbose way and making it works with the option.

### DIFF
--- a/secator/tasks/fping.py
+++ b/secator/tasks/fping.py
@@ -17,9 +17,13 @@ class fping(ReconIp):
 	file_flag = '-f'
 	input_flag = None
 	opts = {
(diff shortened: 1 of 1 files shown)


Moins-2:
My bad I'm stupid. the `--` as opt prefix makes sense for `retry`, `timeout` and `delay`. migth try to fix that when i can

PR #747, opened 2025-11-16
feat(tasks): amass integration

ocervell:
## New Features
* Added Amass-based subdomain enumeration task with support for advanced configuration options including active reconnaissance, ASN data, brute-forcing, DNS resolution settings, and custom wordlists.


ocervell:
@copilot write the actual implementation of amass

ocervell:
## Automated fix applied — flagging for human review (`status-needs-human`)

CI was failing (no fixture; **output parsing was broken end-to-end** — the task declared `output_types=[Tag]`, but `Tag` requires `name`+`match`, which amass JSON lacks, so real runs emitted "Failed to load item as output type" warnings and **zero findings**). Fixes applied to get it green:

**Fixes applied:**
- `output_types` `[Tag]` → `[Subdomain, Ip]`, plus an `on_json_loaded` hook mapping amass `enum -json` (`name` → Subdomain host, `domain` → domain, `sources` → sources, `addresses[].ip` → Ip with IPv4/IPv6 detection).
- Added `tests/fixtures/amass_output.json` + integration inputs/outputs.
- Verified: `secator test unit --tasks amass --test test_tasks` passes (1 Subdomain + 2 Ips, 0 warnings); `secator test lint` clean.

**Needs a human decision:**
1. **No live validation.** The `amass` binary isn't installed in CI, so the mapping was verified against a fixture built from the documented OWASP amass `enum -json` schema, not a live run. Please confirm the field mapping against real amass output (especially `addresses[]` → Ip IPv4/IPv6 handling).
2. **Output scope.** Confirm `[Subdomain, Ip]` is the desired set — `addresses[].cidr`/ASN data is currently dropped; surface it (e.g. as `Record`) if wanted.

PR #886, opened 2026-01-06
fix(tasks): remove x8 since build is broken

ocervell:
## Summary by CodeRabbit

* **Chores**
  * Disabled x8 fuzzer from default URL parameter fuzzing workflows. The arjun fuzzer will now be used as the sole default option for fuzzing operations.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


ocervell:
See https://github.com/Sh1Yo/x8/issues/66 for context

PR #965, opened 2026-04-08
fix(query): support dot notation for nested dict fields in query conditions

### LINKED ISSUE #962: fix(query): no results on nested fields query
e.g:
```
secator r show workflows/236 -q "tag.name == 'technology' and tag.extra_data.product == 'Apache HTTP Server'"
```

will return tags of `name` 'technology' but ignore the `tag.extra_data.product` match (silent error).

ocervell:
Wrap OutputType items in DotMap before adding to the eval context in process_extractor(). This allows queries like `tag.extra_data.product == 'Apache HTTP Server'` to work correctly instead of silently returning no results.

Closes #962

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed condition evaluation to support dot-notation for accessing nested dictionary fields in extractor operations.

### DIFF
--- a/secator/runners/_helpers.py
+++ b/secator/runners/_helpers.py
@@ -1,6 +1,7 @@
 import os
 import re
 
+from dotmap import DotMap
 from secator.output_types import Error
 from secator.utils import deduplicate, debug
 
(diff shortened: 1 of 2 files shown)


ocervell:
@claude will that break workflow templates that use "item.<something>" in their condition field ?

PR #969, opened 2026-04-10
feat(nmap): live results with `-oG -`

ocervell:
## Release Notes

* **New Features**
  * Nmap now displays live progress updates showing percentage completion, elapsed time, and host scan statistics.
  * Added grepable output format option for nmap.
  * Enhanced nmap verbosity control with multiple output levels.

* **Improvements**
  * Progress tracking now uses precise float-based percentages for finer granularity.

### DIFF
--- a/secator/cli.py
+++ b/secator/cli.py
@@ -102,7 +102,10 @@ def workflow(ctx):
 
 
 for config in WORKFLOWS:
-	register_runner(workflow, config)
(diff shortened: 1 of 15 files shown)


ocervell:
Closing - greppable output doesn't have all info that XML has. We need an alternative approach based on reading XML with an interval.

PR #999, opened 2026-04-20
fix(targets): use regex-based hostname validation to detect internal hosts

### LINKED ISSUE #992: fix(targets): some internal hosts are not properly detected as hosts
Internal hosts might have very weird names, and basically any string can be a host.

In autodetect_type, we currently use validators.domain to detect the host part, but this does not catch weird DNS names that can be hosts.

We should consider switching to a regex matching for HOST and HOST_PORT but for the later it will match things we call `slug` today like `redis:3` for grype inputs etc... so we cannot really distinguish between a HOST_PORT and a SLUG in that case...

From the hostname manpage, the definition of a valid host is:
Valid characters for hostnames are ASCII(7) letters from a to z, the digits from 0 to 9, and the hyphen (-). A hostname may not start with a hyphen. Hostnames are often used with network client and server programs, which must generally translate the name to an address for use.

If we really cannot distinguish between a HOST_PORT and a SLUG like `redis:3`, maybe we should just accept HOST_PORT for all the tasks, workflows, or scans currently taking a slug as input.

ocervell:
Replace `validators.domain()` with a regex-based `is_valid_hostname()` that follows the hostname(7) manpage definition.

Fixes detection of multi-label hostnames with non-standard TLDs (`server.internal`, `db.corp`, `web.local`) and single-label hostnames containing a hyphen (`db-server`, `web-01`).

Closes #992

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced hostname validation to recognize internal domain extensions (e.g., `.internal`, `.corp`, `.local`) and hyphenated single-label hostnames.

* **Tests**
  * Expanded test coverage for hostname validation scenarios, including internal TLDs and single-label patterns with port specifications.

### DIFF
--- a/secator/utils.py
+++ b/secator/utils.py
@@ -990,6 +990,34 @@ def format_object(obj, color='magenta', skip_keys=[], predicate=lambda x: True):
 	return ''
 
 
+_HOSTNAME_LABEL_RE = re.compile(r'^[a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?$')
+
+
+def is_valid_hostname(target):
+	"""Check if target is a valid hostname per the hostname(7) manpage definition.
+
+	Accepts multi-label hostnames with any TLD (including non-standard ones like .internal,
+	.corp, .local) and single-label hostnames containing a hyphen (e.g. db-server, web-01).
+	Single-label names without a hyphen (e.g. redis, test123) return False so they can fall
+	through to slug detection and avoid ambiguity with Docker image names and usernames.
+	"""
+	if not target or len(target) > 253:
+		return False
+	try:
+		ascii_target = target.encode('idna').decode('ascii')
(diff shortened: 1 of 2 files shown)


ocervell:
@claude fix coderabbit review comment

ocervell:
@claude Move hostname length validation after IDNA encoding.

Line 1004 validates the hostname length before IDNA encoding, but IDNA can expand Unicode hostnames beyond the 253-byte limit while still passing the subsequent label length checks. For example, a 79-character hostname with 40 'é' labels expands to 319 characters after IDNA encoding, exceeding the RFC 1035 limit but passing the current validation.

Additionally, the fallback at line 1009 (ascii_target = target) allows validation to proceed with Unicode strings when IDNA encoding fails, which may incorrectly accept invalid hostnames. This should return False instead.

Issue #988, opened 2026-04-20
feat(configs): add a way to set options based on a condition


ocervell:
For instance, for a host_recon workflow yaml, we might want to add `skip_host_discovery: true` when `masscan in opts.scanners`. 

The syntax in the template can be like:
```
nmap:
    skip_host_discovery_: true if 'masscan' in opts.scanners else false
```

ocervell:
@claude analyze and implement this

PR #1000, opened 2026-04-20
feat(configs): add conditional options support in workflow YAML

ocervell:
Implements #988

Adds support for setting options conditionally in workflow YAML files using a `key_: {condition: ...}` syntax.

## Release Notes

* **New Features**
  * Added conditional options support to workflows, enabling dynamic task configuration based on runtime conditions. Options now activate automatically when specified criteria are met (e.g., when particular scanners are selected).


ocervell:
@claude I'm not sure that will work in the context of workflows, where options can be prefixed, so the override order matter. Can you investigate on this ?

PR #1001, opened 2026-04-20
fix(runner): exact-match source check prevents prefix-collision task failures

### LINKED ISSUE #987: fix(runner): previous task errors make current task fail
In a workflow, if a previous task has errors, the current task will not run and mark itself as FAILURE.
Desired: it should run anyways and not consider previous errors !

ocervell:
The `startswith(unique_name)` comparison used by `self_errors` (and six other properties) caused tasks with a common name prefix (e.g. `nmap/light` → `nmap_light` and `nmap`) to share each other's errors. The `__iter__` early-exit guard then fired for the shorter-named task (`nmap`) before it ran, reporting a spurious FAILURE.

Fix: add `_is_own_source(source)` that uses exact match + numeric chunk suffix (e.g. `nmap_1`) and correctly rejects same-prefix siblings (`nmap_light`). All seven affected properties updated. Unit tests added.

Closes #987

## Release Notes

* **Bug Fixes**
  * Improved accuracy of result filtering to correctly attribute findings to their source runner, preventing false positives from similarly-named components (e.g., distinguishing between "nmap" and "nmap_light").

* **Tests**
  * Added comprehensive test coverage for source attribution logic, including chunk suffix handling and edge cases.


secator/runners/_base.py
```diff
@@ -297,31 +297,42 @@ def elapsed_human(self):
...
+	def _is_own_source(self, source):
+		"""Return True when *source* was generated by this runner or one of its chunks.
+
+		Accepts exact match (e.g. "nmap") and numeric chunk suffixes (e.g. "nmap_1",
+		"nmap_2") while correctly rejecting same-prefix siblings like "nmap_light".
+		"""
+		if source == self.unique_name:
+			return True
+		prefix = self.unique_name + '_'
+		return source.startswith(prefix) and source[len(prefix):].isdigit()
```
ocervell, line 309: isn't source[len(prefix):] a list ? Would isdigit still work @coderabbit

PR #1005, opened 2026-04-20
fix(exporters): Table exporter broken because of dict change

### LINKED ISSUE #1003: fix(exporters): Table exporter broken because of dict change
Same issue as for the TXT and CSV exporters here: https://github.com/freelabz/secator/pull/994

Using secator r show, the results are dicts instead of output types, so they need to be casted to their respective output types before rendering the table.

@claude please implement the fix

ocervell:
Fix TableExporter to cast dict items to their respective OutputType instances before rendering, matching the fix applied to TXT and CSV exporters in #994.

Closes #1003

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table export functionality to automatically convert dictionary data to appropriate output types, with graceful error handling that preserves original data if type conversion fails.

### DIFF
--- a/secator/exporters/table.py
+++ b/secator/exporters/table.py
@@ -33,7 +33,8 @@ class TableExporter(Exporter):
 							try:
 								item = cls.load(item)
 							except TypeError as e:
-								console.print(f'[bold orange1]Warning:[/] TableExporter could not cast item of type {item.get("_type")!r}: {e}. Skipping.')
+								_type = item.get('_type')
+								console.print(f'[bold orange1]Warning:[/] TableExporter could not cast item of type {_type!r}: {e}. Skipping.')
 								continue
 					cast_items.append(item)
 				if not cast_items:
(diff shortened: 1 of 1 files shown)


ocervell:
@claude Ensure all items are cast to the same type before passing to build_table.

The code only checks items[0] to determine if all items are OutputType instances, but the cast loop can produce mixed lists. If items[0] successfully casts but items[1+] fail the cls.load() call (silent TypeError on line 36), those later items remain as dicts. This causes build_table to fail:

Line 285: operator.attrgetter(*sort_by) requires object attributes; dicts don't have them, causing an AttributeError if sort_by is non-empty.
Line 325: getattr(item, key, '') returns empty strings for dict items instead of their actual values.

Fix: output a warning and skip items that fail to load.

ocervell:
secator/exporters/table.py:36:121: E501 line too long (131 > 120 characters)

@claude fix lint

PR #1006, opened 2026-04-20
fix(query): add --format to report show and improve query expression support

### LINKED ISSUE #963: fix(query): secator report show --format should match format in templates
```
secator r show workflows/236 -q "tag.name == 'technology' and tag.extra_data.get('product') == 'Apache HTTP Server'" --format '{tag.match}-{tag.name}'
```

This should work, but it returns: Invalid output type: {tag}

ocervell:
Fix 'Invalid output type' error and add -- format option to `secator report show` adapted to the new MongoDB-style query architecture.

Closes #963

## Summary by CodeRabbit

* **New Features**
  * Added `--format` option to the `report show` command to format results using custom templates.
  * Query expressions now support Python-style logical operators (`and`/`or`) alongside symbolic operators.
  * Query expressions support Python-style field access patterns for improved readability.

* **Tests**
  * Added comprehensive test coverage for new query and reporting features.

### DIFF
--- a/secator/cli.py
+++ b/secator/cli.py
@@ -959,6 +959,7 @@ def _apply_format(results, fmt):
 			_template = '{' + _field + '}' if _field else None
 
 		if _type not in results:
+			console.print(f'[yellow]Warning: --format type {_type!r} not found in results[/yellow]')
 			continue
 
 		items = results[_type]
@@ -972,6 +973,9 @@ def _apply_format(results, fmt):
 				d = item if isinstance(item, dict) else (item.toDict() if hasattr(item, 'toDict') else {})
 				try:
 					value = _template.format(**{_type: DotMap(d), **d})
+					# DotMap returns an empty DotMap() for missing nested paths; skip such items.
+					if 'DotMap()' in str(value):
+						continue
 					formatted.append(value)
 				except (KeyError, AttributeError):
 					pass
@@ -979,7 +983,7 @@ def _apply_format(results, fmt):
 		else:
 			new_results[_type] = [str(item) for item in items]
 
-	return new_results if new_results else results
+	return new_results
 
 
 @report.command('show')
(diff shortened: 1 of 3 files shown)


ocervell:
1. --format breaks non-console exporters — report.data['results'] items become strings instead of objects.

After _apply_format, items in report.data['results'][type] are plain strings. Downstream exporters fail:

secator/exporters/csv.py line 28 → dict_writer.writerows(items) raises TypeError/AttributeError because strings are not dict-like.
secator/exporters/json.py → Serializes, but structure is invalid (lists of strings instead of objects).
Only secator/exporters/console.py handles this, checking isinstance(item, dict) and falling back to printing.
The default exporter is console, so the default path works, but e.g. secator r show ... -q ... --format '{tag.match}' -o csv fails at runtime.

Options:

Restrict --format to console-only; error early if combined with other exporters.


2. Broaden .get() normalization to handle non-identifier keys.

Line 105 only rewrites identifier-like keys, so extra_data.get('http-title') remains unnormalized and silently queries the wrong field path.

Option: deprecate .get() from queries (remove, no deprecation message needed)


3. Missing nested fields are not skipped — DotMap returns empty sub-maps instead of raising.

The docstring claims items/templates with missing fields are skipped, but DotMap swallows missing attribute access by returning an empty DotMap(). That means for a template like {tag.extra_data.product}-{tag.name}, items lacking extra_data.product won't raise KeyError/AttributeError; they'll produce output like DotMap()-technology instead of being filtered out.

If missing-field filtering is the desired behavior (which the issue #963 example implies — users filter by tag.extra_data.get('product') == 'Apache HTTP Server' and expect formatted output only for matches), you'll want to skip items

4. Silent fallbacks can mask invalid --format specs.

Two behaviors make user errors hard to diagnose:

Line 982 return new_results if new_results else results: if every spec references a type not present in results (typo, empty report), the original unformatted results are returned with no warning — the user sees full OutputType rendering and assumes --format was applied.
Lines 949–962: if the brace regex fails or _type not in results, that spec is silently dropped via continue.
Consider emitting a Warning (via the existing secator.output_types.Warning pattern used elsewhere in this file) when a spec references an unknown type or produces zero formatted items, so users can tell whether the spec didn't match vs. had no data.

Also note: the regex \{(\w+)[.\}] on a template like '{match}' (no type prefix, just a field) treats match as the type and will silently fall through when match isn't in results, which surprises users who write a single-field template without a type qualifier. Consider documenting that {type.field} is required, or handling the unqualified case.

ocervell:
@claude read m'y latest comment and fiw

ocervell:
@claude ditch the .get() altogether, we don't need it

PR #1013, opened 2026-04-21
fix(core): autodetect_type should not check file path if target is too long

### LINKED ISSUE #1012: fix(core): autodetect_target should not check file path if too long
When a target is too long, the Path(target).exists() in secator utils.py will fail with an OSError (File name too long).

@claude please limit that check to target which length is < 255

ocervell:
When a target string is >= 255 characters, `Path(target).exists()` raises `OSError: File name too long` on Linux.

This fix limits the file path existence check to targets shorter than 255 characters.

Closes #1012

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input type detection to properly distinguish between long strings and file paths. The system now correctly classifies inputs exceeding 255 characters as text strings rather than file paths, enhancing type detection accuracy and preventing misclassification of lengthy text data for more reliable input processing.

### DIFF
--- a/secator/utils.py
+++ b/secator/utils.py
@@ -1040,7 +1040,7 @@ def autodetect_type(target):
 		return IBAN
 	elif validators.uuid(target):
 		return UUID
-	elif Path(target).exists():
+	elif len(target) < 255 and Path(target).exists():
 		return PATH
 	elif validators.slug(target):
 		return SLUG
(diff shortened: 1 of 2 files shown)


ocervell:
@claude Verify each finding against the current code and only fix it if needed.

In `@tests/unit/test_utils.py` at line 77, The test case expecting ("a" * 255,
STRING) is wrong because autodetect_type() will call validators.slug on a
255‑character all-'a' string which matches and returns SLUG; update the test in
tests/unit/test_utils.py to either expect SLUG for ("a" * 255) or replace the
test string with a clearly invalid slug (e.g., include a space or punctuation)
and keep the expectation as STRING; ensure the test references
autodetect_type(), and assert against the SLUG/STRING constants accordingly.

PR #1037, opened 2026-04-21
fix(cli): add 'in' operator to query language

### LINKED ISSUE #1035: fix(cli): add 'in' operator to query language
@claude 
secator r show -q "url.status_code in [200,304]" should work with all backends. Implement the necessary conversions for each backend.
secator r show -q "vulnerability.severity in ['high', 'critical']" should also work

ocervell:
Add `in` operator support to secator's query language, so that expressions like `url.status_code in [200,304]` and `vulnerability.severity in ['high', 'critical']` are parsed and converted to MongoDB-style `{$in: [...]} `queries that all backends already handle.

Closes #1035

## Release Notes

* **New Features**
  * Added support for new CLI query operator syntax: `type.field in [val1, val2]` enabling efficient filtering against multiple values in a single query expression.

* **Tests**
  * Introduced comprehensive test coverage for the new query operator, validating support for numeric and string values, including combinations with other filters.


ocervell:
@claude the operator doesn't seem to work with integer fields like url.status_code in [200, 403]

ocervell:
@claude 
In `@secator/query/utils.py` around lines 100 - 101, _IN_RE currently uses
re.IGNORECASE and will match "in [...]" inside quoted strings and before _OP_RE
runs; remove IGNORECASE and make the "in" match only when it appears outside
quotes by reusing the existing quote-tracking logic in _split_logical_op (or
equivalent quote walker): detect the substring " in [" and ensure its index is
not inside a quoted region before applying _IN_RE, or alternatively make the
regex quote-aware; update uses of _IN_RE and any code that depends on its result
(referencing the _IN_RE constant and the _split_logical_op function/_OP_RE flow)
so equality expressions like tag.name == 'values in [x,y]' fall through to the
== branch.

In `@secator/query/utils.py` around lines 146 - 158, The IN-operator branch (m_in
in function handling expressions) silently drops the parsed list when the
left-hand side lacks a field (e.g., "type in [...]") because parts[1] is
missing; update this branch to validate that a field exists after splitting (the
left variable handled where m_in = _IN_RE.match(expr)), and if not, raise a
ValueError (or another explicit exception) describing the malformed "type.field
in [...]" expression so the caller is aware the $in clause was invalid; keep
using _parse_list for values and include reference to _type, field and $in in
the error or validation logic.

ocervell:
@claude do not raise a ValueError. Simply print a console.print(Error(message=''in' operator requires a field (e.g. 'type.field in [...]'): 'url in [200]':)) instead, and ignore the filter.

PR #1041, opened 2026-04-21
fix(cli): stream cmd and description to client in remote worker mode

### LINKED ISSUE #1030: fix(cli): show cmd that was run on remote worker
When in sync=False mode, we should stream an Info object back to the client with the full cmd run (same as print_command and description for sync=True mode)

ocervell:
When sync=False, yield Info objects with the description and full cmd back to the client so they are visible, matching the sync=True behavior where print_description() and print_command() output to the console.

Fixes #1030

## Release Notes

* **New Features**
  * Enhanced task metadata streaming in worker mode to provide improved visibility of task execution, including task descriptions and command information with progress tracking.

### DIFF
--- a/secator/runners/command.py
+++ b/secator/runners/command.py
@@ -474,8 +474,9 @@ class Command(Runner):
 				cmd_str = _s(self.cmd)
 				if self.chunk and self.chunk_count:
 					cmd_str += f' ({self.chunk}/{self.chunk_count})'
+				desc_part = f'{self.description} ' if self.description else ''
 				yield Info(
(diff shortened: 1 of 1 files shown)


ocervell:
I want the message to be one-line and show:
Started task {self.description} ([bold gold3]{self.unique_name}[/]) (cmd=[dim white]{self.cmd}[/])

ocervell:
@claude see my latest comment and fix

ocervell:
@claude  If no self.description, task shouldn't be in parenthesis and the description should not span an extra space because it's empty.

ocervell:
@claude  without description it should be: Started task [bold gold3]{task.name}[/] (no parenthesis around task name)

PR #1044, opened 2026-04-21
feat(driver): add JSON driver for live report stream

### LINKED ISSUE #1019: feat(json driver): turn json exporter into json driver
Currently we have to wait till the end of a runner scan to get the results when working with the json exporter (-o json), which means while the runner is running we cannot dig into results using `secator r show` and have to wait till the runner ends instead.

We had the idea that instead of writing the JSON report at the end, we could turn the JSON exporter into a driver with hooks/ instead,
with a file `hooks/json.py` where:
- update_runner would update the 'info' key in the JSON report
- update_item would update the 'results.<output_type>' key in the JSON report

However, JSON reports can get huge, and we don't want to load everything in memory for each update, so we should find a solution that doesn't involve reading the entire results.<output_type> list to append a new item.

@claude please come up with a few solutions and we'll choose the best one.

ocervell:
Closes #1019

Implements Option C from the design discussion: turns the JSON exporter into a real-time driver so `secator r show` returns results while a run is in progress.

**How it works:**
- On each item, the driver appends one JSON line to `report.<type>.jsonl` (O(1), no reads)
- Runner state is written to `report.meta.json` on init/interval
- `JsonBackend` falls back to sidecar files when `report.meta.json` exists but `report.json` does not
- On `on_end`, sidecars are merged into the final `report.json` (same format as before) and cleaned up

No new dependencies (stdlib only). Fully backward-compatible with existing `report.json` files.

## Summary by CodeRabbit

* **New Features**
  - JSON output driver is now available.
  - JSON reports can be generated for scans, workflows, and tasks, including per-type sidecar outputs that are merged into a single report.
  - Querying now supports both completed and live (in-progress) report formats.

* **Improvements**
  - Report listing and the CLI now surface live/running reports and optionally include sub-tasks when requested (e.g., via "show all").


ocervell:
@claude the 'secator r list' command doesn't show currently running runners, even though we now have the report.meta.json available : can you investigate why and fix it ?

ocervell:
@claude In `@secator/hooks/json.py` around lines 74 - 83, The current loop that reads
sidecar_path uses a single try/except around the whole file read so a
json.JSONDecodeError on one corrupt line aborts the rest; change it so
I/O/opening errors remain caught by the outer try/except but each
json.loads(line) is wrapped in its own try/except (catch json.JSONDecodeError
and optionally log via debug(f'Bad JSON in {sidecar_path}: {e}',
sub='hooks.json')) and continue to the next line, ensuring valid lines still get
appended to items before assigning results[type_name] = items.

ocervell:
```
secator r show
``` 

is not showing any results while the scan is running.

@claude please help fix that.

ocervell:
Closed in favor or #1312

PR #1063, opened 2026-04-27
fix(templates): support nested dot notation in field extractor

### LINKED ISSUE #1062: fix(templates): nested json queries for field should work
In YAML templates for workflows / scans, the following field dot notation should work correctly:

```
pw_task:
  description: Check password is valid
  targets_:
     - type: tag
       field: extra_data.password
```

You can look at the query module to check how we fixed this issue there, and apply the same to process_extractors.

@claude please address this issue.

ocervell:
Fixes #1062

Add `_format_nested()` helper in `process_extractor` to resolve `{extra_data.password}`-style dot notation by traversing nested dicts, matching the pattern used in `secator/query/json.py`'s `get_nested_field()`.

Previously, `str.format(**item.toDict())` was used, which fails for nested fields because Python's format strings attempt attribute access on dicts rather than dict key traversal.

## Summary by CodeRabbit

* **New Features**
  * Added support for nested field references using dot-notation in template formatting, allowing users to access nested data with notation like `{nested.subfield}`.
  * Improved robustness of formatting operations with graceful handling of missing nested values—now returns empty strings instead of errors when referenced fields are unavailable.

### DIFF
--- a/secator/runners/_helpers.py
+++ b/secator/runners/_helpers.py
@@ -10,7 +10,7 @@ def _format_nested(template, data):
 	"""Format a string template supporting nested dot notation like {extra_data.password}.
 
 	Replaces {key} and {key.subkey} tokens by traversing nested dicts.
-	Raises KeyError for missing or non-traversable keys so misconfigured templates surface loudly.
+	Missing or non-traversable keys resolve to empty string.
 	"""
 	def replace_token(match):
 		key = match.group(1)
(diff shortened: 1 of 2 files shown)


ocervell:
@claude In `@secator/runners/_helpers.py` around lines 9 - 26, _fix_nested currently swallows missing keys in replace_token leading to silent empty-string
substitutions; change it to raise a KeyError when traversal fails so
misconfigured templates surface loudly. In function _format_nested (and its
inner replace_token) detect when value becomes None for a requested dotted key
(key variable) and raise KeyError(f"Missing placeholder: {key}") instead of
returning '' — keep the final re.sub(r'\{([\w.]+)\}', replace_token, template)
unchanged so replace_token is the source of the failure semantics.

ocervell:
@claude actually, forget about the KeyError since it will make the extractors fail if the value is None (which can happen). However, when None or '' is returned, do not add that to the current targets (we don't want empty or None targets or option sets)

PR #1064, opened 2026-04-27
fix(workflows): use scan reports dir for storing results

### LINKED ISSUE #1061: fix(workflows): use scan reports dir for storing results
Workflows currently store their results in their own directories even when they are part of a scan, when they should save everything to the scan reports_dir instead.

@claude please fix.

ocervell:
Workflows currently store their results in their own directories even when they are part of a scan. This fix passes the scan's `reports_folder` to child workflows so all results are saved to the scan's directory.

Fixes #1061

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reports folder path handling for workflow execution to ensure proper report destination assignment.

### DIFF
--- a/secator/runners/_base.py
+++ b/secator/runners/_base.py
@@ -286,9 +286,9 @@ class Runner:
 
 	@property
 	def file_name(self):
-		"""Filesystem-safe name for output files. Uses node_id when running inside a workflow/scan to avoid conflicts when the same task appears in multiple workflows."""
+		"""Filesystem-safe name for output files. Uses node_id when inside a workflow/scan to avoid conflicts."""
 		if self.config.node_id:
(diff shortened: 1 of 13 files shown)


ocervell:
@claude for a domain scan, we have a 'testssl' in domain_recon, and another in subdomain_recon. Will that mean their testssl.json output file will conflict ? Can you propose a solution ?

ocervell:
@claude okay, make the fix above, but make sure it doesn't impact anything else (how we track tasks etc...) since self.unique_name is used in a lot of places.

ocervell:
@claude also replace any '/' by '_' in the node_id so there is no need to create subdirectories.
Also, fix the lint:
secator/runners/_base.py:289:121: E501 line too long (164 > 120 characters)

PR #1071, opened 2026-05-04
fix(scan): subsequent workflow tasks not defaulting to original targets

### LINKED ISSUE #1070: fix(scan): subsequent workflow tasks in scan not defaulting to original targets
When we have a scan definition, for instance like:

```yaml
type: scan
name: test
workflows:
  workflow1:
  workflow2:
    targets:
      - type: port
        field: '{host}:{port}
```

and the workflow2 defined like:

```yaml
type: workflow
name: workflow2
tasks:
  task1:
  task2:
```

`task` will take the right targets, but `task2` will be skipped like it has no targets.

@claude identify why and propose a fix in a branch.
Please note that:
* use the same workflow directly outside of the context of the scan) will work fine, and both tasks receive the input targets correctly.
* adding `targets_: - target.name` to the task2 options doesn't work either, when enabling debug I get:
```
[DBG] extractor     extracted results (ancestor_id: workflow2, key: targets):
[DBG] extractors    extracted 0 / 54 for key targets with extractor <DYNAMIC(target.name)> (ancestor_id: workflow2)
[DBG] extractors    computed_inputs
``` --> maybe the ancestor_id is not set correctly for targets passed by a scan to it's children workflow ?
* the problem might also lies in process_extractors code, but this code handles all target filtering so make careful changes to it only if you have to and analyze the blast radius of any change made.

### PR DESCRIPTION
**Pull request by ocervell**
Fix for issue #1070.

Subsequent tasks in a scan workflow were being skipped because `forwarded_opts` (containing the `targets_` extractor from the scan config) was only applied to the first task in the Celery chain. All subsequent tasks received empty inputs and no extractor, causing them to be skipped.

The fix applies `forwarded_opts` to ALL tasks when `chain_previous_results=True`. The existing `ancestor_id` mechanism in `add_result` correctly labels chained results so subsequent tasks can find them via the extractor.

Closes #1070

Generated with [Claude Code](https://claude.ai/code)) | [View job run](https://github.com/freelabz/secator/actions/runs/25308980959

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workflow chaining to consistently apply forwarded options to all tasks in the workflow, ensuring extractors and other parameters are properly applied throughout the entire chain execution.

### DIFF
--- a/secator/celery.py
+++ b/secator/celery.py
@@ -297,15 +297,16 @@ def mark_runner_started(results, runner, enable_hooks=True):
 	for item in results:
 		runner.add_result(item, print=False)
 
-	# For workflow runners inside a scan (has_parent=True), emit scope-tagged Targets so that
-	# ALL tasks in this workflow can query the same target pool via parent_scope, regardless of
+	# For workflow runners inside a scan (has_parent=True), emit ancestor-tagged Targets so that
+	# ALL tasks in this workflow can query the same target pool via ancestor_id, regardless of
 	# whether they have their own targets_ extractor or not.
 	# This must happen after adding prior results so that scan-level target extractors (e.g.
 	# targets_: [{type: port, …}]) resolve correctly against the accumulated results.
 	if runner.has_parent and getattr(runner.config, 'type', None) == 'workflow':
 		from secator.runners._helpers import run_extractors
 		from secator.output_types import Target as TargetType
-		scope = runner.config.name
+		parent_ancestor = runner.ancestor_id  # scan name (set by scan.py)
+		scope = f'{parent_ancestor}.{runner.config.name}' if parent_ancestor else runner.config.name
 		target_dynamic_opts = {k: v for k, v in runner.dynamic_opts.items() if k.rstrip('_') == 'targets'}
 		if target_dynamic_opts:
 			ctx = {'ancestor_id': runner.ancestor_id, 'node_chain_start': True}
@@ -314,9 +315,9 @@ def mark_runner_started(results, runner, enable_hooks=True):
 			scoped_inputs = runner.inputs
 		for t in scoped_inputs:
 			target = TargetType(name=t)
-			target._context['scope'] = scope
+			target._context['ancestor_id'] = scope
 			runner.add_result(target, print=False)
-		debug(f'Runner {runner.unique_name}: emitted {len(scoped_inputs)} scope-tagged targets (scope={scope})', sub='celery')
+		debug(f'Runner {runner.unique_name}: emitted {len(scoped_inputs)} ancestor-tagged targets (ancestor_id={scope})', sub='celery')
 
 	# Run mark_started
 	runner.mark_started()
--- a/secator/celery.py
+++ b/secator/celery.py
@@ -297,6 +297,27 @@ def mark_runner_started(results, runner, enable_hooks=True):
 	for item in results:
 		runner.add_result(item, print=False)
 
+	# For workflow runners inside a scan (has_parent=True), emit scope-tagged Targets so that
+	# ALL tasks in this workflow can query the same target pool via parent_scope, regardless of
+	# whether they have their own targets_ extractor or not.
+	# This must happen after adding prior results so that scan-level target extractors (e.g.
+	# targets_: [{type: port, …}]) resolve correctly against the accumulated results.
+	if runner.has_parent and getattr(runner.config, 'type', None) == 'workflow':
+		from secator.runners._helpers import run_extractors
+		from secator.output_types import Target as TargetType
+		scope = runner.config.name
+		target_dynamic_opts = {k: v for k, v in runner.dynamic_opts.items() if k.rstrip('_') == 'targets'}
+		if target_dynamic_opts:
+			ctx = {'ancestor_id': runner.ancestor_id, 'node_chain_start': True}
+			scoped_inputs, _, _ = run_extractors(runner.results, target_dynamic_opts, runner.inputs, ctx=ctx)
+		else:
+			scoped_inputs = runner.inputs
+		for t in scoped_inputs:
+			target = TargetType(name=t)
+			target._context['scope'] = scope
+			runner.add_result(target, print=False)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by ocervell** [comment]
@claude Your fix is incorrect and won't work, because the issue is a bit deeper.

What should happen:
- task1 runs extractors based on dynamic opts passed by the workflow (so it correctly resolves the right targets), and then using add_result will add Target object with _context['ancestor_id'] = 'workflow2'
- task2 runs extractors (with added ancestor_id = 'workflow2') and should be able to correctly resolve the right targets.

Currently it seems that Target objects are not emitted when targets computed with extractors, and it makes sense since _run_extractors() run after adding targets to the runner in _base.py. After computing inputs with _run_extractors we should output Target objects as well.

Actually I think I know what the fix should be:
1. Always add ancestor_id to task context (workflow.py line 106):
```py
task_opts['context']['ancestor_id'] = current_id
```
3. Also add a new context attribute named `node_chain_start` (workflow.py line 107):
```py
task_opts['context']['node_chain_start'] = (ix == 0 or parent_ix == 0)
```

4. In runners/_base.py, Move the target add_result after the self._run_extractors() run (e.g line 209):
```py
# Add targets to results
targets = [Target(name=target) for target in self.inputs]
for target in targets:
	self.add_result(target, print=False, output=False)
	self.debug(f'added target {str(target)} (ancestor_id: {target._context.get("ancestor_id")})')
```

5. In _base.py (line 541), add the 'node_chain_start' to the ctx object:
```py
ctx = {..., 'node_chain_start': self.context.get('node_chain_start', False)}
```

7. In runners/_helpers.py (line 168), add:
```py
node_chain_start = ctx.get('node_chain_start')
```
8. and line 179, change the ancestor condition:
```
if ancestor_id and not node_chain_start:
```

That should make sure that:
- the first task in a workflow / scan always queries targets without an ancestor_id
- the first task outputs targets with the right ancestor_id
- subsequent tasks in a workflow query targets scoped to the ancestor_id, which now exist

ocervell:
@claude Okay, this works if we add:
    # targets_:
    # - type: target
    #   field: name
to task2, but if we omit it it should work as well (query all current targets with ancestor_id)

ocervell:
@claude actually remove the last fix:
```
elif not node_chain_start and ancestor_id:
# No targets_ extractor defined for this chained task — default to all Target
# objects tagged with the ancestor_id so subsequent tasks get their inputs
# automatically without requiring an explicit extractor definition.
ancestor_targets = deduplicate([
	item.name for item in results
	if item._type == 'target' and item._context.get('ancestor_id') == str(ancestor_id)
])
if ancestor_targets:
	debug('defaulted to ancestor targets', obj=ancestor_targets, sub='extractors')
	inputs = ancestor_targets
```
this is unecessary, but you'll also need to restore to the original when building workflow:
```
if (ix == 0 or parent_ix == 0) and forwarded_opts:
  task_opts.update(forwarded_opts)
```
otherwise the task will try to run the same query as passed to the workflow, with ancestor_id: workflow2, which will not work.

ocervell:
@claude Adding no `targets_` block to our task2 should mean: use the targets provided to the workflow (in scan context they can be dynamic), so propose a way we could do this (don't implement the fix yet)

ocervell:
@claude you are right about your cavea, a workflow defined in a scan context like:

```
type: scan
name: test
workflows:
  workflow1:
  workflow2:
    targets:
      - type: port
        field: '{host}:{port}
```

and the workflow2 defined like:

```
type: workflow
name: workflow2
tasks:
  task1:
    targets_:
      - type: target
        field: name
        condition: "'test' in item.name"
  task2:
```

I think this will skip running task1 when run with target scanme.nmap.org (because no target has 'test' in name), although it **should** actually run task2 with 'scanme.nmap.org'.

So I think a task marked with `node_chain_start` should also emit filtered targets from the workflow filter, before the actual target filtering by the `targets_` key.

That puts another issue into light:
- When task1 has it's own targets_ filter, will that overwrite the workflow targets_ filter in scan contexts ? Please investigate and propose a fix.

==> My current understanding of the current flow (make sure to verify this)
- scan starts (N targets)
- workflow 1 inits, builds Celery workflow (no targets_)
- workflow 2 inits, builds Celery workflow (targets_ but resolved to [] since workflow 1 hasn't started yet)
- scan Celery workflow built
- workflow 1 starts, runs mark_runner_started, emit Target object with ancestor_id=None
- workflow 2 starts, runs mark_runner_started, emit Target object with ancestor_id=None
- task 1 starts, queries Target with ancestor_id = None, emit Target objects with ancestor_id='workflow2'
- task 2 starts, queries Target with ancestor_id = 'workflow2'
**==> so if task 1 doesn't emit any filtered target, task 2 will be skipped (unwanted).**

==> The ideal flow:
- scan starts (N targets)
- workflow 1 inits, builds Celery workflow (no targets_)
- workflow 2 inits, builds Celery workflow (targets_ but resolved to [] since workflow 1 hasn't started yet)
- scan Celery workflow built
- workflow 1 starts, runs mark_runner_started, emit Target object with scope='test.workflow1' (node id) and ancestor_id='test'
- workflow 2 starts, runs mark_runner_started, emit Target object with scope='test.workflow2' (node id) and ancestor_id='test'
- task 1 starts, queries Target with scope 'test.workflow2', emit Target objects with scope='test.workflow2.task1'
- task 2 starts, queries Target with scope 'test.workflow2', emit Target objects with scope='test.workflow2.task2'
**=> This way:
- Workflow / task targets are separated
- No need to pass the dynamic filter to the first target
- No need to have complex if ancestor_id things in extractors

Please think, argue if you don't agree, figure out what gotchas it could have, and if everything checks out, implement my fix.

ocervell:
@claude some key observations to fix:
- Let's keep the name ancestor_id instead of parent_scope, no need to have both as they represent the same thing: we just need to make sure that it's set properly at every step throught the scan > workflow > task chains.
- Non-target types (port, url, …) can use the same ancestor_id-filtering as targets for sure.
- We should use full identifiers for ancestor_id, e.g: <scan_name>.<workflow_name>.<task_name> to avoid any potential conflict that would lead to querying wrong targets (e.g: two tasks have the same name in different workflows)

PR #1075, opened 2026-05-05
fix(target-filtering): secondary task should not take ancestor results as direct input

### LINKED ISSUE #1074: fix(target-filtering): secondary task in workflow should not take results as direct input
PR #1073 introduced nice improvements but there is one unwanted behavior:
- When a task in a workflow produces a result, it gets automatically passed as target to the next task if that task has no input extractors: it shouldn't as it might break some tasks.

So we should:
- Remove the code block (line 99 in runners/_helpers.py)
```
# if not node_chain_start and ancestor_id:
# 	ancestor_results = [
# 		str(item) for item in results
# 		if item._context.get('ancestor_id') == str(ancestor_id) and type(item) not in _meta_types
# 	]
# 	combined = deduplicate(scoped_targets + ancestor_results)
# else:
```

and just keep the content of the else.

- Remove the test `test_non_first_task_includes_ancestor_id_results` in `unit/test_target_filtering.py`

@claude address and make a PR

ocervell:
Remove the ancestor_id-based fallback that caused secondary workflow tasks without input extractors to receive all ancestor results as direct targets. Now only scope-tagged Target objects are used as fallback inputs.

Also removes the test `test_non_first_task_includes_ancestor_id_results` that validated the unwanted behavior.

Fixes #1074

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified target combination logic when parent scope filtering is enabled. Input targets now derive exclusively from scope-matched results with consistent de-duplication.

* **Tests**
  * Updated target filtering test cases to align with revised scope-based target selection behavior.

### DIFF
--- a/secator/runners/_helpers.py
+++ b/secator/runners/_helpers.py
@@ -44,15 +44,11 @@ def run_extractors(results, opts, inputs=None, ctx=None, dry_run=False):
 		inputs = []
 	if ctx is None:
 		ctx = {}
-	# Pull parent_scope and node_chain_start from opts when not in ctx
-	# (_run_extractors in _base.py only adds ancestor_id to ctx)
 	if 'parent_scope' not in ctx:
 		ctx['parent_scope'] = opts.get('parent_scope')
 	if 'node_chain_start' not in ctx:
 		ctx['node_chain_start'] = opts.get('node_chain_start', False)
 	parent_scope = ctx.get('parent_scope')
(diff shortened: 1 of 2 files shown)


ocervell:
@claude fix lint:
secator/runners/_helpers.py:54:2: F841 local variable 'ancestor_id' is assigned to but never used
secator/runners/_helpers.py:55:2: F841 local variable 'node_chain_start' is assigned to but never used

PR #1077, opened 2026-05-06
feat(tree): prune condition-failing nodes from scan/workflow tree display

ocervell:
## Summary

- Adds `prune_runner_tree(tree, opts, inputs)` to `secator/tree.py` that walks the runner tree bottom-up and removes nodes whose `if` conditions evaluate to `False` against the current opts/targets
- Wires it into `log_start()` in `_base.py` so the \"Scan built:\" / \"Workflow built:\" message only shows tasks and workflows that will actually run
- Execution paths in `workflow.py` and `scan.py` are untouched — this is display-only

## Test Plan

- [ ] Run `secator test unit` — 302 tests pass, 0 failures
- [ ] Run `secator test lint` — clean
- [ ] Run a scan or workflow that has conditional tasks with options that exclude some tasks — verify the excluded tasks do not appear in the \"built:\" tree
- [ ] Run the same scan/workflow without the excluding options — verify the full tree is shown

## Summary by CodeRabbit

* **New Features**
  * Scan/workflow tree displays are pruned before rendering, hiding nodes whose conditions evaluate to false for the current options and inputs; CLI tree output reflects this behavior and preserves nodes on evaluation errors.

* **Documentation**
  * Added a formal design and implementation plan for the pruned tree display and wiring steps.

* **Tests**
  * Added unit tests covering pruning behavior, condition evaluation edge cases, group handling, and input-driven pruning.


ocervell:
@claude In `@secator/cli_helper.py` around lines 275 - 277, prune_runner_tree is being
given a raw click argument (tree_opts.get('inputs', [])) which may be a string
not a list, causing incorrect pruning; before calling prune_runner_tree after
build_runner_tree, normalize the inputs into a proper list (use the existing
expand_input utility if available, or convert a single string into [inputs] or
[] when None) and pass that list instead of the raw value so prune_runner_tree
sees a list of targets; update the call site that currently does
prune_runner_tree(tree, tree_opts, tree_opts.get('inputs', [])) to compute
normalized_inputs = expand_input(tree_opts.get('inputs')) or ( [v] if
isinstance(v, str) else v or [] ) and pass normalized_inputs to
prune_runner_tree.

PR #1085, opened 2026-05-12
fix(worker): prevent target duplication when running in worker mode

### LINKED ISSUE #1069: fix(worker): target is duplicated when running in worker
When running a simple task in a worker, it gets duplicated and the remote worker runs two targets instead of one:

e.g:
secator worker 2>&1 &
secator x httpx github.com

In worker logs:
[INF] Loaded 2 targets for httpx
       github.com (host)
       github.com (host)
httpx-toolkit -irh -srd /home/jahmyst/.secator/reports/local/tasks/15/.outputs -u github.com -json -sr -rstr 100000 -rsts 100000 -header 'User-Agent:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36' -threads 50

the cmd still ends up resolving only one target because we have deduplication, but there are still 2 targets passed for some reason.

@claude please fix.

ocervell:
Fixes #1069

When a task is dispatched via Celery (start_runner), both the Task runner and the command runner create Target objects for the same inputs. Since they share the same unique_name, self_targets counts both, reporting "Loaded 2 targets" instead of 1.

Fixed by checking for existing Target names in prior results before creating new Target objects from inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where duplicate target result entries were being created unnecessarily during runner initialization. The system now checks for existing target entries before instantiating new ones, eliminating redundant data and ensuring cleaner result sets.

[Review Change Stack](https://app.coderabbit.ai/change-stack/freelabz/secator/pull/1085)


ocervell:
@claude we don't want to loop through all results for such a simple task, as it can take a long time for 100k+ objects.

What we should do instead is skip adding re-adding targets for tasks when they are run with the Task runner.

Issue #1092, opened 2026-05-20
fix(query): exception when regex starts with '*' or when it's a number


ocervell:
```
$ secator r show -q "vulnerability.id ~= '*CVE-2026-28780'"
                         __
   ________  _________ _/ /_____  _____
  / ___/ _ \/ ___/ __ `/ __/ __ \/ ___/
 (__  /  __/ /__/ /_/ / /_/ /_/ / /
/____/\___/\___/\__,_/\__/\____/_/     v0.32.0

                        freelabz.com

Traceback (most recent call last):
  File "/usr/local/bin/secator", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "/root/.secator/lib/python3.11/site-packages/rich_click/rich_command.py", line 404, in __call__
    return super().__call__(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.secator/lib/python3.11/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.secator/lib/python3.11/site-packages/rich_click/rich_command.py", line 187, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/root/.secator/lib/python3.11/site-packages/click/core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.secator/lib/python3.11/site-packages/click/core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.secator/lib/python3.11/site-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.secator/lib/python3.11/site-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.secator/lib/python3.11/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
[...]

PR #1105, opened 2026-06-02
fix(query): prevent exception when regex starts with '*' or is a number

ocervell:
Fix two related bugs in the query subsystem (issue #1092).

**Bug 1 — `re.error: nothing to repeat` when pattern starts with `*`**
The `$regex` lambda in `json.py` passed the pattern directly to `re.search()` with no error handling. Fixed by replacing the lambda with `_regex_match()` which catches `re.error` and returns `False`.

**Bug 2 — `TypeError` when pattern is a number**
`_parse_value()` converts numeric-looking strings to `int`/`float`, which breaks `re.search`. Fixed in `_parse_single_expr` by skipping coercion when the operator is `$regex`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved regex query operator robustness: invalid regex patterns no longer cause errors and evaluate gracefully instead
  * Fixed regex pattern preservation in query expressions to ensure patterns are correctly processed
  * Enhanced handling of regex queries with unexpected data types to prevent crashes


ocervell:
@claude when the regex starts with a '*', just strip that '*' as it will work fine that way.

PR #1107, opened 2026-06-02
fix(output_types): update vuln to not compare cvss score + vector + epss

bpce-it-dis-soc-sof:
## Summary by CodeRabbit

* **Refactor**
  * Updated vulnerability data structure to improve internal handling of scoring fields during data comparisons.


ocervell:
That was indeed preventing duplicate checks to work properly. Thanks !

PR #1137, opened 2026-06-07
feat(profile): support workspace, drivers and exporters in profiles

ocervell:
## Summary

Profiles can now set `workspace`, `drivers` (list) and `exporters` (list) in addition to `opts`, so a profile can fully describe where results go, which drivers wire up, and how output is exported — not just task/workflow/scan option overrides.

```yaml
type: profile
name: example
enforce: false
workspace: my-workspace
drivers:
  - mongodb
exporters:
  - json
  - csv
opts:
  ...
```

## Semantics

All three fields honor the existing `enforce` flag, consistent with `opts`:

- **workspace** (scalar): enforced → overrides the user's workspace; non-enforced → applied only if the user kept the default workspace.
- **drivers** (list): additive. Hooks are loaded through the existing dynamic-driver mechanism (`discover_external_drivers` → `import_dynamic` → `register_hooks`) and added to `context['drivers']`, so **worker mode re-loads them automatically** via `Runner.__setstate__` (the #1126 path). Also merged into `self._hooks` so they propagate to sub-runners. Unsupported / addon-missing drivers emit a `Warning` and are skipped.
- **exporters** (list): enforced → replaces user/config exporters; non-enforced → unions with them (written back to `run_opts['output']`, consumed by the existing exporter resolution).

The "Loaded profile ..." log line now also shows the new fields.

## Implementation

All changes are contained in `Runner.resolve_profiles()` (`secator/runners/_base.py`) plus a small `_apply_profile_drivers()` helper. No profile schema change is needed (profiles are `TemplateLoader`/`DotMap`). No shipped profiles were modified.

## Known limitation

A profile that *adds* the `api` driver does not trigger the CLI-side workspace-name API lookup (`get_workspace_name`), since that only runs in `cli_helper.py`.

## Testing

End-to-end with an httpx task and a dropped `~/.secator/templates/test.yml` profile changing all three fields:

- workspace → reports written under `reports/test_workspace/...`
- drivers → custom external driver loaded and its hooks fired on both the Task and the httpx sub-runner
- exporters → `json` + `csv` reports produced
- precedence verified: `-ws cli-override` beat the non-enforced profile workspace; `-o txt` unioned to `txt, json, csv`

Adds 8 unit tests (workspace apply/block/enforce, exporter union/replace, driver load+register, unsupported-driver skip).

- `secator test unit --test test_runners` → 45 passed
- `secator test unit --test test_celery` → 16 passed (pickle/worker path unaffected)
- `secator test lint` → clean

## Summary by CodeRabbit

* **New Features**
  * Profiles now support workspace selection with configurable precedence levels (default, user-set, enforced).
  * Profiles now include driver and exporter configuration options.

* **Bug Fixes**
  * Improved remote information output handling in piped execution mode.

* **Tests**
  * Added unit tests for profile-driven configuration, covering workspace, driver, and exporter behavior.

### DIFF
--- a/secator/cli.py
+++ b/secator/cli.py
@@ -897,13 +897,29 @@ def profile(ctx):
 
 @profile.command('list')
 def profile_list():
-	table = Table()
+	table = Table(highlight=True)
 	table.add_column('Profile name', style='bold gold3')
 	table.add_column('Description', overflow='fold')
+	table.add_column('Enforced', justify='center')
+	table.add_column('Workspace', overflow='fold')
+	table.add_column('Drivers', overflow='fold')
+	table.add_column('Exporters', overflow='fold')
 	table.add_column('Options', overflow='fold')
 	for profile in PROFILES:
-		opts_str = ', '.join(f'[yellow3]{k}[/]=[dim yellow3]{v}[/]' for k, v in profile.opts.items())
-		table.add_row(profile.name, profile.description or '', opts_str)
+		opts_str = ', '.join(f'[yellow3]{k}[/]={v}' for k, v in profile.opts.items())
+		enforced_str = '[bold red]✓[/]' if profile.enforce else ''
+		workspace_str = profile.workspace or ''
+		drivers_str = ','.join(profile.drivers) if profile.drivers else ''
+		exporters_str = ','.join(profile.exporters) if profile.exporters else ''
+		table.add_row(
+			profile.name,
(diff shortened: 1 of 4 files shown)


ocervell:
@coderabbit can you check if this way the workspace will be enabled at the right time ? I know the base runner creates workspace folder, we want to make sure that happens **after** setting the workspace option from profiles.

PR #1144, opened 2026-06-08
feat(cli): dynamic install addons

### LINKED ISSUE #1143: feat(cli): dynamic install addons
In #1116 we added dynamic drivers and exporters loading.

For some of these, we might need addons to install required dependencies.

E.g: if we implement a `elasticsearch` driver, we might need to install the ElasticSearch Python SDK.

In that case, we need a way to run: `secator install addons elasticsearch` and define the addons dependencies somewhere else than in the `pyproject.toml`.

Find a way to: dynamically load a dependency file in the ~/.secator/templates folder named addons.json that will contain something like:
```
{
   'elasticsearch': {
        'pypi_dependencies': [
            'elasticsearch<10'
        ],
        'next_steps': []  # next steps for `secator install addons elasticsearch` command
    },
   'other_dep': {
        'install_cmd': 'curl https://other_dep.com/[install_version]',
        'install_version': '2.3.4'
        'install_github_handle': 'other/dep',
        ... and all the other install fields currently supported by the `Command` runner
   }
}
```

Same as for tasks /workflows / scans templates, drivers, and exporters, we should dynamically load the `addons.json` file and dynamically add the commands to Secator CLI:
* `secator install addons elasticsearch` --> will run `{sys.executable} -m pip install elasticsearch < 10`
* `secator install addons other_dep` --> will run the same workflow as for `secator install tools <tool_name>`, so in this case: 1. GithubInstaller on other/dep --> 2. if not available, SourceInstaller with curl https://other_dep.com/2.3.4

@claude make this happen.

ocervell:
Implements dynamic loading of addon install commands from `~/.secator/templates/addons.json` as requested in #1143.

## Changes
- `secator/loader.py`: new `load_external_addons()` — reads and caches addons.json
- `secator/cli.py`: `_install_external_addon()` and `_load_external_addon_commands()`
- `tests/unit/test_loader.py`: 7 unit tests for `load_external_addons`

Closes #1143

## Release Notes

* **New Features**
  * Added support for installing external addons via dedicated CLI commands (e.g., `secator install addons <name>`)
  * Installation process supports both package managers and custom tool installers
  * Post-installation guidance automatically displayed upon successful setup
  * Offline mode prevents addon installation to ensure consistency

### DIFF
--- a/secator/cli.py
+++ b/secator/cli.py
@@ -2114,36 +2114,29 @@ def install_ai():
 
 def _install_external_addon(name, config):
 	"""Install an external addon defined in addons.json."""
-	import shlex
-	from secator.installer import SourceInstaller, ToolInstaller
+	from secator.installer import ToolInstaller
 
 	if CONFIG.offline_mode:
 		console.print(Error(message='Cannot run this command in offline mode.'))
 		sys.exit(1)
 
-	pypi_deps = config.get('pypi_dependencies', [])
 	next_steps = config.get('next_steps', [])
-
-	if pypi_deps:
-		quoted_deps = ' '.join(shlex.quote(dep) for dep in pypi_deps)
-		cmd = f'{sys.executable} -m pip install {quoted_deps}'
-		status = SourceInstaller.install(cmd)
-	else:
-		tool_attrs = {
-			'__name__': name,
-			'install_pre': config.get('install_pre', None),
-			'install_post': config.get('install_post', None),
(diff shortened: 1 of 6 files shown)


ocervell:
@claude great, but actually `pypi_dependencies` can actually be integrated in `ToolInstaller` as we might need it for other tasks (e.g: a `PythonRunner` derived task that would need an extra package to run like `pandas`).

PR #1168, opened 2026-06-11
fix(nuclei): uppercase CVE id

bpce-it-dis-soc-sof:
Always run upper() on the `cve_id` in the `nuclei` template as some templates have the CVE id in lowercase.

## Summary by CodeRabbit

* **Bug Fixes**
  * CVE identifiers now display in uppercase format.


ocervell:
Thanks for the PR ! 
Good catch, not sure how this was skipped in our integration tests. Merging !

PR #1169, opened 2026-06-11
fix(cli): put loader success messages in debug instead of console.print

bpce-it-dis-soc-sof:
The console.print are a little too obnoxious when many external integrations are made, so it's preferable to put them in `debug`.

Also, it can impeed the auto-completion.

## Summary by CodeRabbit

* **Chores**
  * Enhanced logging consistency for external plugin loading. Debug-level messages now properly surface during successful initialization of external tasks, drivers, and exporters, improving visibility for development and troubleshooting workflows. Error and validation logging behavior remains unchanged.


ocervell:
@bpce-it-dis-soc-sof Thank you ! You are totally right, after testing it seems to be preventing proper auto-completion.

PR #1181, opened 2026-06-15
feat: make the api driver work end-to-end in the CLI (queries, workspaces, operators)

ocervell:
Makes the `api` driver (Secator Cloud) usable end-to-end from the CLI — via `--driver api` or `drivers.defaults: api` — and fixes the query correctness bugs found along the way. All verified against live prod data (workspace `T2S`).

## Features
- **`secator ws use <name>` creates the workspace remotely** when the api driver is active (idempotent — an existing workspace of the same name is reused). Adds `addons.api.workspace_create_endpoint` + `hooks.api.create_workspace()`.
- **New query operators**: `!~=` (not matching regex → `{$not:{$regex}}`) and `not in [..]` (→ `$nin`), wired through the CLI parser and the json + sqlite backends (mongo/api handle them natively).

## Query fixes
- **`query` / `report show` ignored `drivers.defaults`**: built the engine context from the raw `--driver` (empty when unset), so it resolved the workspace via the API but queried the local JSON backend. Now uses the *resolved* backend.
- **`report list -ws <name>` returned nothing over the api driver**: the API filters runners by `workspace_id` (ObjectId); the command passed the raw name. Now resolves name→id (`r list` with no `-ws` already worked).
- **Findings with no `is_false_positive` field were all dropped**: the enforced base query used exact `is_false_positive: False`, which doesn't match a missing field — so `q -ws T2S port` returned **0** despite **2163** findings. Changed to `{$ne: True}`.
- **Compound `&&` on the same field collapsed**: a `dict.update()` merge dropped earlier clauses (`host !~= a && host != b` kept only `!= b`). Now same-field disjoint operators merge into one field dict; otherwise a clean **top-level `$and`** of full fragments (MongoDB has no field-level `$and`, and a *mixed* `{field, $and:[..]}` dict made `validate_query_fields` silently drop sibling keys — which had also been triggering the default `_type:{$nin:[target,ai]}` exclusion).

## Tests
- Updated the 4 base-query assertions to `is_false_positive: {$ne: True}`.
- Isolated the report-delete e2e tests from the developer's `drivers.defaults` (force `--driver local`): `resolve_backend()` reads the real `secator.config.CONFIG`, not the mocked `cli.CONFIG`, so a dev with `drivers.defaults: api` was routing local-driver deletes to the API and failing (CI was unaffected).

## Style
- Cheatsheet (`secator cs`) comments now render in a real muted color (`grey42`) instead of the ANSI faint `dim` attribute, which terminals/less ignore (so under `LESS=-R` the colors showed but the dim didn't). Comments are now uniformly dim; command coloring is unchanged.

## Verification
Live `T2S` workspace: `q` now returns the expected counts (e.g. `port.host !~= 'cluster026..' && port.host !~= 'smtp'` → 165 = 2163 − 1008 − 990). Full query/report unit suite green (177 passed); the 4 originally-failing CI tests pass.


ocervell:
Added two more commits while testing against live data:

**5. New query operators `!~=` (not matching regex) and `not in [..]`**
- `!~=` → `{$not: {$regex: ...}}`, `not in [a, b]` → `$nin`, wired through the CLI parser (`query/utils.py`) and the json + sqlite backends (mongo/api handle them natively).

**6. Fixed compound `&&` queries on the same field**
- The AND merge used `dict.update()`, so two clauses constraining the same field collapsed — `port.host !~= 'cluster026..' && port.host != 'smtp..'` kept only the `!=` clause. Now same-field operator dicts are merged (`{$not:.., $ne:..}`); unmergeable clauses fall back to `$and`.

Verified live (T2S): that compound query now returns **165** = 2163 all ports − 1008 (cluster026) − 990 (smtp); previously the `!~=` half was silently dropped.

ocervell:
@claude fix the CodeRabbit review comments

PR #1199, opened 2026-06-19
feat(celery): cap worker-loss redeliveries to abandon repeatedly-killed tasks

ocervell:
## Problem

When a worker is killed mid-task — a cgroup **OOMKill** of the child process, or a node **memory-pressure eviction** — the task never returns a result. With `task_acks_late` + `task_reject_on_worker_lost`, the broker redelivers the task (same Celery id) to a fresh worker. That's the desired recovery, **except** for a task that OOMs on *every* run (e.g. a memory spike the inner `task_memory_limit_mb` monitor doesn't catch in time): it loops forever — OOM → redeliver → OOM — and the surrounding chord/workflow can **never complete**.

This was hit in prod: a headless `katana` task on `worker-large` was evicted (`The node was low on resource: memory`), the Job failed (`BackoffLimitExceeded`), and the workflow stalled.

## Change

Add **`celery.task_max_retries`** (`-1` = disabled). `run_command` counts redeliveries (keyed by the stable Celery id) and, once the cap is exceeded, **abandons** the task instead of re-running it: it returns a forwarded result list with a `FAILURE` `Error` appended (`"Task <name> abandoned after N retries (worker repeatedly lost — likely OOM kill or node eviction)."`), so the chord proceeds and the workflow finishes cleanly — mirroring the existing inner memory-limit warning.

### Backend-agnostic counter
The counter lives on the **Celery result backend's generic key/value interface**, not Redis internals and not any Secator data backend (Mongo/Postgres/SQLite):
- atomic `incr` when the backend provides it (Redis, Memcached) — race-free;
- `get`/`set` read-modify-write otherwise (filesystem, S3, GCS, cache, …) — safe because same-task redeliveries are **sequential**;
- backends with neither (database/RPC) degrade to *cap disabled* (no crash).

## Safety / rollout
**Inert by default** — `task_max_retries=-1` and `task_acks_late=False` ship unchanged, so this commit changes no behavior. Enabling the cap requires, at deploy time:
- `task_acks_late=1` + `task_reject_on_worker_lost=1`,
- `task_max_retries` set (e.g. `3`),
- ⚠️ **`broker_visibility_timeout` raised above the max task lifetime** (tasks run up to `activeDeadlineSeconds=10800`); leaving it at the default 1h would duplicate-execute long tasks once `acks_late` is on.

## Tests
`tests/unit/test_celery.py` — get/set fallback (real filesystem backend), atomic-incr path, unsupported-backend degradation, and the abandon path returning a `FAILURE` error. Full file: 22 passing.

## Scope note
This fully covers the **child-OOMKill retry loop**. A **whole-pod eviction** kills the parent too, so its redelivery comes via the slower Redis visibility-timeout restore; that case is primarily addressed by task-profile sizing (see #1198) + Guaranteed QoS. Externally un-sticking an already-stuck chord is a separate follow-up (Layer 3).

## Summary by CodeRabbit

* **New Features**
  * Added configurable worker-loss retry limit for tasks. When a task exceeds the maximum allowed retries after worker loss, it is automatically abandoned and marked as failed rather than retried indefinitely.

### DIFF
--- a/secator/celery.py
+++ b/secator/celery.py
@@ -15,7 +15,7 @@ from retry import retry
 from secator.celery_signals import setup_handlers
 from secator.definitions import IN_WORKER
 from secator.config import CONFIG
-from secator.output_types import Info, Target as TargetOutput
+from secator.output_types import Error, Info, Target as TargetOutput
 from secator.rich import console
 from secator.runners import Scan, Task, Workflow
 from secator.runners._helpers import run_extractors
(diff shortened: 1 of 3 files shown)


ocervell:
Pushed the review fixes + the L1 delivery flag (commit `634b2591`):

- **Off-by-one** fixed — extracted `worker_loss_retries_exhausted(delivery_count, max_retries)`; the initial delivery no longer counts as a retry, so `task_max_retries=3` now allows the initial run + 3 redeliveries. Unit-tested at the boundary.
- **Key expiry** — worker-loss counter keys get a best-effort TTL (`result_expires`) on both the atomic-incr and get/set paths, so they no longer accumulate forever.
- **L1 flag** — added `worker_cancel_long_running_tasks_on_connection_loss` (default `False`) wired into `app.conf`, alongside `task_acks_late`/`task_reject_on_worker_lost`. Inert until enabled in deployment.

Deferred: the terminal-doc no-op guard (a redelivered task whose runner doc is already `done`) depends on stable runner identity (`on_build`, PR #1202) and is ineffective without it — it'll land once #1202 merges. `TestWorkerLossRetryCap` is 6/6 green; full `test_celery.py` 23/23.

ocervell:

PR #1203, opened 2026-06-21
feat(celery): finalize in-flight task on worker eviction so the chord proceeds

ocervell:
## Problem

On a **K8s pod eviction** (`SIGTERM` → grace → `SIGKILL`) the worker running a task dies. With `task_acks_late` on the **Redis** broker, its message is only redelivered after the broker **visibility timeout** — which we derive *above* each pool's deadline (hours, on the long-task pool). So the surrounding chord/workflow **stalls for that entire window**.

This is the one OOM/eviction case the other layers don't cover:

| Scenario | Recovery |
|---|---|
| Task timeout / memory soft-limit | secator monitor self-stops → partial results (existing) |
| Fast-spike OOM (child killed, **master survives**) | `task_reject_on_worker_lost` requeues immediately (~18s), capped by #1199 |
| **Whole-pod eviction (master dies too)** | ❌ nothing requeues fast on Redis → **visibility-timeout stall** |

A whole-pod eviction has no surviving master, so `reject_on_worker_lost` can't help and Redis falls back to the visibility timeout.

## Fix

Catch the `worker_shutting_down` signal and raise a flag the running task's **monitor thread** polls. On shutdown it stops the task early via the existing `stop_process(exit_ok=True)` path — the same self-stop already used for the timeout and memory-limit cases — so the task **returns its partial results through the normal completion path** and the chord proceeds **in seconds** instead of waiting for the visibility timeout.

- **`celery_signals.py`** — `SHUTDOWN_FLAG` + `worker_shutting_down_handler` (wired unconditionally); stale flag cleared on worker boot. File-based so it crosses the prefork master→child boundary.
- **`command.py`** — the monitor checks the flag, and its poll interval is capped (`MONITOR_POLL_SECONDS=5`) so an eviction is caught well within the pod's `terminationGracePeriodSeconds` (60s).

Goes through Celery's normal completion path, so the chord is satisfied the ordinary way — no pool change (stays prefork), no Celery upgrade, no `revoke` (which yields `REVOKED`/`TaskRevokedError`, the wrong tool here — that's for *cancel*, e.g. `stop_runner`).

## Tests

- `test_shutdown_flag_lifecycle` — flag set/clear.
- `test_monitor_stops_running_command_on_shutdown` — a real `sleep 30` command stops early (well under 30s) and emits the eviction `Warning`.

Validated end-to-end locally: evicting a worker mid-chord, the **chord callback fired in ~5s** (vs. the ~hours visibility timeout) with partial results from the evicted member.

Part of the OOM/eviction robustness set alongside #1199 (worker-loss retry cap) and #1202 (on_build runner identity).

## Summary by CodeRabbit

* **New Features**
  * Implemented graceful worker shutdown that terminates running commands and automatically saves partial results, preventing indefinite task hangs during system shutdown
  * Enhanced command monitoring to responsively detect shutdown signals and exit long-running tasks quickly

* **Tests**
  * Added comprehensive unit tests validating worker shutdown eviction and command termination behavior

### DIFF
--- a/secator/celery_signals.py
+++ b/secator/celery_signals.py
@@ -36,9 +36,10 @@ def clear_shutdown_flag():
 def worker_shutting_down_handler(**kwargs):
 	"""Raise the eviction flag so the in-flight task's monitor stops it early and returns."""
 	try:
+		SHUTDOWN_FLAG.parent.mkdir(parents=True, exist_ok=True)
 		SHUTDOWN_FLAG.write_text("1")
-	except Exception:
-		pass
+	except OSError as e:
+		console.print(Info(message=f'Failed to raise worker shutdown flag: {e}'))
 
 
 def get_lock_file_path():
--- a/secator/celery_signals.py
+++ b/secator/celery_signals.py
@@ -15,6 +15,31 @@ IDLE_TIMEOUT = CONFIG.celery.worker_kill_after_idle_seconds
 STATE_DIR = Path("/tmp/celery_state")
 STATE_DIR.mkdir(exist_ok=True, parents=True)
 
+# Eviction flag. On worker shutdown (e.g. a K8s pod SIGTERM eviction) we raise this flag; the
+# running task's monitor thread (in the prefork child, a separate process) polls it via a file
+# and stops early, returning partial results so the surrounding chord proceeds — instead of the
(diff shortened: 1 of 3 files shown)


ocervell:

PR #1222, opened 2026-06-25
feat(ai): record total billed tokens per ai run

ocervell:
## Goal

Add per-run AI-token accounting to the `ai` task so the cloud platform can bill/quota it. The `ai` task already gets real billed token usage **per LLM call** (`call_llm` → `response.usage.total_tokens` + litellm `completion_cost`) but never aggregated it. This persists the **per-run total** on the runner so the billing chore can read it — the AI analog of how `context.scan_hours` is used for run-hours billing.

## What changed

- **`secator/tasks/ai.py`**
  - `_init_options` seeds `context["ai_tokens"]` (int) and `context["ai_cost"]` (float) so the field lands on the task doc even with zero LLM calls.
  - `_account_usage(usage)` sums a single `call_llm` usage dict onto `self.context`. Missing/None/malformed usage counts as 0 and never raises.
  - Wired in after the **main loop** `call_llm` (before any empty-response `continue`, so each billed call is counted once) and after the **intent-detection** `call_llm`.
  - `_drain_history_usage()` rolls billed usage accrued by history summarization into `context.ai_tokens` once per iteration.
- **`secator/ai/history.py`** — `ChatHistory.compact()` now records its summarization call's billed usage on `billed_tokens`/`billed_cost` (the task drains it). Missing usage = 0.

`self.context` is copied onto every item's `_context` and persisted to MongoDB, so **`context.ai_tokens`** lands on the task doc. Subagent/batch `ai` tasks are separate runners with their own task doc + own `context.ai_tokens`, so the chore sums across docs without double-counting. Consistent in chat and attack modes.

## The field the billing chore reads

**`context.ai_tokens`** (cumulative billed tokens, int). `context.ai_cost` (float) is also persisted.

## Tests

New `tests/unit/test_ai_tokens.py` (9 tests, all green): N-call sum, missing/None usage = 0, malformed usage doesn't crash, exact persisted key, summarization drain-once, `compact()` records usage / handles missing usage, and two end-to-end `_run_loop` runs asserting `context.ai_tokens` == sum (and 0 with no usage). flake8 clean.

## Summary by CodeRabbit

* **New Features**
  * Added tracking of billed AI token and cost usage during runs.
  * Run summaries now include accumulated usage from both direct model calls and history compaction.

* **Bug Fixes**
  * Improved handling of missing or invalid usage data so accounting stays accurate.
  * Ensured history-related usage is counted only once and then cleared from temporary storage.

* **Tests**
  * Added coverage for usage tracking, invalid usage values, and end-to-end run accounting.

### DIFF
--- a/secator/ai/history.py
+++ b/secator/ai/history.py
@@ -128,6 +128,8 @@ class ChatHistory:
     # (history summarization/compaction). The owning `ai` task drains these into
     # context.ai_tokens so summarization is billed alongside the main loop.
     billed_tokens: int = 0
+    billed_prompt_tokens: int = 0
+    billed_completion_tokens: int = 0
     billed_cost: float = 0.0
 
     def add_system(self, content: str) -> None:
@@ -400,6 +402,14 @@ class ChatHistory:
         usage = result.get("usage") or {}
         try:
             self.billed_tokens += int(usage.get("tokens") or 0)
+        except (TypeError, ValueError):
+            pass
+        try:
+            self.billed_prompt_tokens += int(usage.get("prompt_tokens") or 0)
+        except (TypeError, ValueError):
+            pass
+        try:
+            self.billed_completion_tokens += int(usage.get("completion_tokens") or 0)
         except (TypeError, ValueError):
             pass
         try:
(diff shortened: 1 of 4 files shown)


ocervell:

PR #1232, opened 2026-06-29
feat(runners): redact sensitive task options from serialized/printed state

ocervell:
## What

Adds a `sensitive: True` option meta-flag (a sibling of the existing `internal` flag). An option value flagged sensitive is masked with `[REDACTED]` everywhere runner state is **persisted, serialized, or printed**, while the real value still travels in-flight to the worker and runs in the actual subprocess.

This lets users pass **their own addon API keys** through the existing options channel — e.g. the `ai` task's `api_key` (now marked sensitive) — without the secret ever landing in the Mongo runner doc, API responses, `--driver api` egress, or logs.

## Why

We want users to bring their own secrets (BYO addon keys, etc.) per run. The opts channel already carries these to the worker and persists runner state — but it persisted the secret in plaintext. Rather than build a separate config-override channel, we keep secrets on the opts channel and make the **data model** secret-aware. (Design: `docs/superpowers/specs/2026-06-29-sensitive-options-and-per-run-overrides-design.md` in `gke-admin`.)

## Leak surfaces covered

1. **`toDict()` `run_opts`** — Mongo runner doc, API responses, `--driver api` egress (`_base.Runner.resolved_opts`).
2. **The built command string** — a secret passed as a CLI flag (`--api-key X`) is masked in the `cmd` that `toDict()` and `print_command` expose, while the executed subprocess keeps the real value.
3. **Debug / console echoes** of the command and its options.

## Accepted, non-user-facing exposures (out of scope, by design)

- The value still rides the Redis broker message transiently (same as the existing external-keys vault).
- If a task passes a secret as a CLI flag, it appears in worker process args (`ps`); tasks that read the secret from an env var avoid this.

## Changes

- **`runners/_base.py`** — `Runner.sensitive_opt_names` (resolved from the runner's own `opts`/`meta_opts` or, for `Task`/`Workflow`/`Scan`, from its config tree's task classes; memoized) + redaction in `resolved_opts`. New `REDACTED_OPT_VALUE` constant.
- **`runners/command.py`** — build a redacted command alongside the real one; emit it from `toDict()` and `print_command`. **No behavior change for tasks without a sensitive opt** (`cmd_redacted == cmd`, flag off).
- **`cli_helper.py`** — drop the `sensitive` key before building click options (it's not a click param).
- **`tasks/ai.py`** — mark `api_key` `sensitive: True`.
- **`tests/unit/test_sensitive_opts.py`** — both leak surfaces (run_opts + cmd) for a Command task, the `Task`-runner path (ai `api_key`), and the non-sensitive no-op path.

## Testing

- New `test_sensitive_opts.py`: 7 passing.
- `test_command.py`, `test_python_runner.py`, `test_ai_task_opts.py`: green (33 total).
- `flake8` clean on all changed files.
- Verified `secator x ai --help` still builds the `--api-key` option (the `sensitive` key no longer breaks click).
- Manually verified: executed cmd keeps the real secret; `toDict()`/printed cmd and `run_opts` are masked.

## Summary by CodeRabbit

* **New Features**
  * Sensitive option values are now automatically hidden in displayed and serialized command/runner output.
  * Secret fields such as API keys are marked to stay masked wherever they appear.

* **Bug Fixes**
  * Prevented sensitive option metadata from being treated like a regular command option.
  * Improved command printing so secret values are no longer exposed in logs or debug output.

### DIFF
--- a/secator/runners/_base.py
+++ b/secator/runners/_base.py
@@ -1,4 +1,3 @@
-import copy
 import gc
 import json
 import logging
@@ -333,28 +332,6 @@ class Runner:
 			logger.warning('Could not resolve sensitive option names for %s (redaction may be incomplete): %s', getattr(self, 'unique_name', self), e)  # noqa: E501
 		self._sensitive_opt_names_cache = names
 		return names
-
-	@staticmethod
-	def _redact_sensitive(obj, sensitive):
-		"""Recursively mask sensitive option values in a serialized structure (in place).
-
-		Masks the ``default``/``value`` fields of any option-conf dict keyed by a sensitive
-		opt name, and any scalar held directly under a sensitive opt name. Callers must pass
-		a COPY — never the live config — since this mutates ``obj``.
-		"""
-		if isinstance(obj, dict):
-			for k, v in obj.items():
-				if k in sensitive:
-					if isinstance(v, dict):
-						for field in ('default', 'value'):
-							if v.get(field):
-								v[field] = REDACTED_OPT_VALUE
-					elif v and not isinstance(v, (list, dict)):
-						obj[k] = REDACTED_OPT_VALUE
-				Runner._redact_sensitive(v, sensitive)
-		elif isinstance(obj, list):
-			for v in obj:
-				Runner._redact_sensitive(v, sensitive)
 
 	@property
 	def resolved_opts(self):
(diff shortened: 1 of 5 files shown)


secator/runners/_base.py
```diff
@@ -284,9 +288,58 @@ def _process_config(self, config):
...
+				except ValueError:
+					# Unknown task name in this node: skip it, but keep scanning the rest.
+					continue
+		except Exception as e:
+			# Discovery failed for the whole tree: we can't enumerate sensitive opts, so
+			# redaction may be incomplete. Surface it loudly rather than failing silently
+			# (a silent miss here would let a secret through into serialized/printed state).
+			logger.warning('Could not resolve sensitive option names for %s (redaction may be incomplete): %s', getattr(self, 'unique_name', self), e)  # noqa: E501
+		self._sensitive_opt_names_cache = names
+		return names
```
ocervell, line 321: This code is way too complicated for just getting sensitive option names. Aim to simplify it. I don't think there is a need for a broad try / except here, or caching.

ocervell: Done in b71f123e. Dropped the memoization (`discover_tasks()` is already `@cache`d, so it was redundant) and both `try/except` layers. The exception-based `get_task_class` lookup is replaced by a non-raising registry lookup — unknown task names just skip, and a direct Command/PythonRunner instance is already covered by `collect(self)`. Net −22/+9, and it now fails *closed* (the old broad `except` swallowed discovery errors → could return an incomplete set and leak a secret). Tests: test_sensitive_opts + test_runners green.

PR #1240, opened 2026-06-30
feat(vulnerability): add status field + carry-over across re-scans

ocervell:
## Summary

Adds a `status` field to the `Vulnerability` finding type (`NEW` | `ACKNOWLEDGED` | `FIXED`, default `NEW`) and makes it carry over across re-scans. This is the **secator core** layer (§1) of the Vulnerability status feature.

**Repo 1/3 of Vulnerability status (secator → secator-api → secator-ui).**

## The field (`output_types/vulnerability.py`)
- New `status: str = field(default='NEW', compare=False)` — `compare=False` so it does **not** affect dedup equality; identity stays `name`/`id`/`matched_at`.
- `__post_init__` normalizes/validates: empty / `None` / unknown → `'NEW'`, uppercased; allowed set is `NEW` / `ACKNOWLEDGED` / `FIXED` (`Vulnerability.STATUSES`).
- Added `STATUS` to `_table_fields` so it renders; added `STATUS = 'status'` to `definitions.py`.

## Carry-over behavior
- Added `'status'` to `duplicate_main_copy_fields` defaults of both `MongodbAddon` and `SqliteAddon` (`config.py`).
- `hooks/_dedup.py` (`compute_duplicate_updates`): introduced a small `_is_unset(field, value)` helper. For `status` it treats `''` / `None` / `'NEW'` as **unset**; all other fields keep the generic `not value` emptiness check. Applied to both the previous-value guard and the current-value copy condition, so a prior `ACKNOWLEDGED`/`FIXED` is copied forward onto a re-found main whose status is still the default `NEW`, while a never-touched vuln stays `NEW`.

## Tests
- `tests/unit/test_output_types.py` (`TestVulnerabilityStatus`): default `NEW`; empty/`None`/unknown coerce to `NEW`; valid values preserved + uppercased + trimmed; status does **not** affect equality or `_compare_key()`.
- `tests/unit/test_dedup.py` (`TestComputeDuplicateUpdates`): prior `ACKNOWLEDGED` copies onto a new `NEW` main; a new `FIXED` main is not overwritten; a prior `NEW` is not carried; non-status fields keep `not value` semantics.

All 59 relevant unit tests pass; `flake8 secator/` is clean (matches CI `secator test lint`). Pre-existing `test_config.py` failures are an unrelated env-fixture issue (`SECATOR_DIRS_DATA`), present on `origin/main` without these changes.

## Summary by CodeRabbit

* **New Features**
  * Added vulnerability status tracking with supported states like `NEW`, `ACKNOWLEDGED`, and `FIXED`.
  * Status values are now normalized automatically for consistent display and handling.

* **Bug Fixes**
  * Improved duplicate-item handling so status information carries forward correctly without overwriting finalized statuses.
  * Prevents empty or placeholder status values from being treated as real updates.

---
_Recreates #1209 (merged in 0d7bf7d, then reverted by #1234). Same changes, re-applied on top of current main._


secator/output_types/vulnerability.py
```diff
@@ -40,7 +43,8 @@ class Vulnerability(OutputType):
 	_duplicate: bool = field(default=False, repr=True, compare=False)
 	_related: list = field(default_factory=list, compare=False)
 
-	_table_fields = [MATCHED_AT, SEVERITY, CONFIDENCE, NAME, ID, CVSS_SCORE, TAGS, EXTRA_DATA, REFERENCE]
+	_table_fields = [MATCHED_AT, SEVERITY, CONFIDENCE, NAME, ID, CVSS_SCORE, STATUS, TAGS, EXTRA_DATA, REFERENCE]
+	STATUSES = ('NEW', 'ACKNOWLEDGED', 'FIXED')
```
ocervell: This should be a global def in constants.py, not here

secator/output_types/vulnerability.py
```diff
@@ -66,6 +70,11 @@ def __post_init__(self):
 		self.severity_nb = severity_map.get(self.severity, 6)
 		self.confidence_nb = severity_map[self.confidence]
 
+		# Normalize status: coerce empty / None / unknown values to 'NEW', uppercase.
+		# Allowed values are NEW / ACKNOWLEDGED / FIXED (see STATUSES).
+		status = (self.status or '').strip().upper()
+		self.status = status if status in self.STATUSES else 'NEW'
```
ocervell: Those two lines aren't needed, statuis default is 'NEW' already

secator/hooks/_dedup.py
```diff
@@ -1,6 +1,20 @@
 # secator/hooks/_dedup.py
 
 
+def _is_unset(field, value):
```
ocervell: remote his. Unneeded,complicates code. Just treat status the same way we treat other fields.

PR #1241, opened 2026-06-30
feat(ai): AI-task reliability, robustness & security hardening (27 fixes)

ocervell:
AI-task **reliability / robustness / security** hardening — the full remediation of the
2026-06-30 deep review (`docs/superpowers/specs/2026-06-30-ai-task-reliability-review.md`).

**27 / 29 findings fixed**, one focused PR per finding, merged here across 7 rounds.
H8 excluded (overlaps `feat/sensitive-options`); H3 deferred to a design decision (event-driven
human-wait — see the review doc's H3 proposal).

### 🔴 Critical
- **C1** #1239 — sanitize LLM-supplied subagent opts (deny control keys + force `dangerous=False`)
- **C2** #1238 — wrapper-aware shell guardrail (stop exec-wrapper command laundering)
- **C3** #1254 — `celery_id` idempotency marker; redelivery no longer replays/re-bills a turn

### 🟠 High
- **H1** #1235 — bound retries on persistent 429 (no infinite loop)
- **H2** #1244 — tool-pair-aware trim/compact (no orphan-tool 400s)
- **H4** #1248 — cap subagent recursion depth + per-turn fan-out
- **H5** #1249 — persist pending docs for plain-chat & max-iter remote turns
- **H6** #1238 — fix `rm -rf /*`-class destructive-command deny
- **H7** #1237 — correlate permission prompts by `prompt_uuid` (no stale auto-approve)
- **H9** #1243 — "allow this command" is a true one-shot, not a session rule
- **H10** #1242 — fail closed when guardrail ask-prompts are exhausted

### 🟡 Medium
- **M1** #1256 — cap unbounded shell output into history
- **M2** #1252 — don't silently spawn persistence-less subagents on empty drivers
- **M3** #1255 — model-window-aware history trim budget
- **M4** #1251 — 400s fail fast (keep orphan-tool repair)
- **M5** #1247 — token-count fallback when LLM `usage` is absent
- **M6** #1246 — default-safe target/path guardrail dimensions (no fail-open on missing catch-all)
- **M7** #1260 — gate privileged `add_finding` types (block injected trusted-target minting)
- **M8** #1253 — normalize encoded IPs before SSRF/metadata deny
- **M9** #1257 — treat `curl -o`/`wget -O` output paths as writes
- **M10** #1245 — channel hygiene: answer/timeout race + stale-pending GC
- **M11** #1264 — extensible exec-wrapper peeling (closes the C2 laundering class)
[...]

PR #1284, opened 2026-07-06
feat(celery): cap worker-loss redeliveries to abandon repeatedly-killed tasks

ocervell:
**Re-extracted from the closed #1199** — this generic Celery worker-reliability fix was folded into the `ai-resiliency` branch (#1241, AI-task hardening, base `canary`) and its standalone PR closed. It's general worker-pool infra (nothing AI-specific), so it belongs on `main` on its own timeline. Cherry-picked clean onto current `main`.

## What it does
Under `task_acks_late` + `task_reject_on_worker_lost` (prod config), a task that dies with its worker (SIGKILL / OOM / node eviction) is redelivered with the **same Celery id**. A task that OOMs on *every* run would loop forever (OOM → redeliver → OOM), blocking its chord indefinitely. This caps redeliveries: `run_command` counts worker-loss redeliveries (Redis `celery-task-meta-worker-loss-<id>`); once `task_max_retries` is exceeded it **abandons** the task with a clear `abandoned after N delivery attempts` result so the chord/workflow can finish. `task_max_retries=-1` disables the cap (default). Also adds the connection-loss cancel flag.

## Validation
Exercised end-to-end earlier (prefork child-loss → requeue → abandon at cap, 6/6). **`tests/unit/test_celery.py`: 23 passed** on current `main`; flake8 clean.

## Note
`feature:worker-reliability`. Companion PR extracts the eviction-finalize half (was #1203). Prod workers run `-P prefork -c 1`, so the fast child-loss path is live.

## Summary by CodeRabbit

* **New Features**
  * Added a limit for repeated worker-loss retries so long-running jobs can stop retrying after a configurable cap.
  * Introduced an option to cancel in-flight tasks when the broker connection is lost, helping prevent stuck background work.

* **Bug Fixes**
  * Improved handling of interrupted tasks so workflows can complete with a clear failure result instead of hanging indefinitely.
  * Added safer retry counting to better track redelivered tasks across worker losses.

### DIFF
--- a/secator/celery.py
+++ b/secator/celery.py
@@ -15,7 +15,7 @@ from retry import retry
 from secator.celery_signals import setup_handlers
 from secator.definitions import IN_WORKER
 from secator.config import CONFIG
-from secator.output_types import Info, Target as TargetOutput
+from secator.output_types import Error, Info, Target as TargetOutput
 from secator.rich import console
 from secator.runners import Scan, Task, Workflow
 from secator.runners._helpers import run_extractors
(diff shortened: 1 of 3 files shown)


ocervell:
Addressed the CodeRabbit findings (commit pushed):
- **Attach context before abandon** — `opts['context'] = context` is now set before the early `abandon_task` return, so an abandoned task keeps `celery_id`/`worker_name`/`routing_key` even when `opts` had no context.
- **Gate on `task_reject_on_worker_lost`** — the cap now requires `task_acks_late` **and** `task_reject_on_worker_lost` (the setting that actually re-queues on abrupt worker death).
- **`task_max_retries` validation** — `Field(default=-1, ge=-1)` so values `< -1` can't enable the cap and abandon on first delivery.
- **Expire diagnosability** — `_expire_worker_loss_key` now only swallows `AttributeError`/`NotImplementedError`; other backend errors are debug-logged.

`tests/unit/test_celery.py` 23/23, flake8 clean.

PR #1285, opened 2026-07-06
feat(celery): finalize in-flight task on worker eviction so the chord proceeds

ocervell:
**Re-extracted from the closed #1203** — like the cap PR (#1284), this generic worker-reliability fix was folded into the `ai-resiliency` branch (#1241, AI-task hardening, base `canary`) and its standalone PR closed. It's general infra (applies to every runner, not AI), so it belongs on `main` independently. Cherry-picked clean onto current `main` (CI-retrigger noise commits dropped).

## What it does
On a **graceful worker shutdown** (SIGTERM — k8s pod eviction / node drain / rolling update), `worker_shutting_down` sets a flag and the running task's monitor thread **stops the task early and finalizes it** — so the chord's `mark_runner_completed` still fires and the workflow proceeds, instead of the task hanging until the broker visibility timeout (hours on the long pool). Works on any pool (monitor-thread based). Per-task flag is cleared at start so a stale flag can't finalize a fresh task.

## Validation
Exercised end-to-end earlier (SIGTERM → task stopped ~9s vs ~60s, chord proceeds, 5/5). **`tests/unit/test_eviction.py`: 3 passed; `test_celery.py`: 17 passed** on current `main`; flake8 clean.

## Note
`feature:worker-reliability`. Companion: #1284 (worker-loss cap). #1202 (on_build identity) is already on `main` (in the pending 0.40.0).

## Summary by CodeRabbit

* **New Features**
  * Added graceful shutdown handling for long-running workers so running jobs can detect eviction and stop sooner.
  * Improved monitoring responsiveness during shutdown checks to reduce the chance of stale work continuing after termination begins.

* **Bug Fixes**
  * Cleared stale shutdown state at startup to prevent old shutdown signals from affecting new runs.
  * Added safeguards so interrupted tasks emit a warning and terminate cleanly instead of hanging.

### DIFF
--- a/secator/celery_signals.py
+++ b/secator/celery_signals.py
@@ -36,9 +36,10 @@ def clear_shutdown_flag():
 def worker_shutting_down_handler(**kwargs):
 	"""Raise the eviction flag so the in-flight task's monitor stops it early and returns."""
 	try:
+		SHUTDOWN_FLAG.parent.mkdir(parents=True, exist_ok=True)
 		SHUTDOWN_FLAG.write_text("1")
-	except Exception:
-		pass
+	except OSError as e:
+		console.print(Info(message=f'Failed to raise worker shutdown flag: {e}'))
 
 
 def get_lock_file_path():
--- a/secator/celery_signals.py
+++ b/secator/celery_signals.py
@@ -15,6 +15,31 @@ IDLE_TIMEOUT = CONFIG.celery.worker_kill_after_idle_seconds
 STATE_DIR = Path("/tmp/celery_state")
 STATE_DIR.mkdir(exist_ok=True, parents=True)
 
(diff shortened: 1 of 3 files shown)


ocervell:
**Parked for later** (owner decision). Concern: not enough confidence in the in-flight runner's behavior on worker shutdown — what happens to the running Python runner / external tool process, partial results, and how finalize-early interacts with redelivery. Too many edge cases to merge without more validation.

Safe to park: this only helps the **graceful SIGTERM** (k8s eviction / drain / rolling update) path — a *hard* worker death (SIGKILL/OOM) is handled by acks_late redelivery + the worker-loss cap (#1284) and the broker visibility-timeout fallback. So the platform works correctly without this; it just finalizes slower on graceful eviction. Revisit when we can validate the runner/partial-results semantics end-to-end.

PR #1295, opened 2026-07-08
fix(command): authoritative sudo required

bpce-it-dis-soc-sof:
Some commands might use sudo in their options set, like for running RCEs with custom pentest tools.

E.g: 
```
zabbix_get -k 'system.run[sudo -l]'
```

Currently, that would mean the system autodetect sudo in the cmd string, and pre-prends 'sudo' to the command, which is not the desired outcome in that case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sudo password prompting so it now follows the command’s actual sudo requirement instead of checking for the word “sudo” in the command text.
  * Reduced unnecessary password prompts when sudo is not required, while keeping prompts enabled when security settings call for them.

### DIFF
--- a/secator/runners/command.py
+++ b/secator/runners/command.py
@@ -502,9 +502,8 @@ class Command(Runner):
 			self.print_command()
 
 			# Check for sudo requirements and prepare the password if needed
-			sudo_required = re.search(r'\bsudo\b', self.cmd)
 			sudo_password = None
-			if CONFIG.security.prompt_sudo_password and sudo_required:
+			if CONFIG.security.prompt_sudo_password and self.sudo_required:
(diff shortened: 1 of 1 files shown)


ocervell:
Thanks for the PR !

@claude please check if any task, tool installs, or addon installs, or raw Command.execute() blocks could be impacted by this change.

ocervell:
@claude it was actually `self.requires_sudo`. Fixed. Can you check if that impacts anything ?

ocervell:
@claude we do need to handle these cases by passing explicit `requires_sudo` attribute to the `Command.execute` (especially for package / source / tools installs as we don't want to break those, they are sensitive).

PR #1297, opened 2026-07-10
fix(query): support {'_id':{'$in':ids}, **filters} fan-in queries (RC#6 prereq)

ocervell:
## Why

RC#6 (fan-in Python-heap OOM): a `_base` runner receives the **full accumulated result set** and materializes it to compute its targets via `run_extractors`/`process_extractor`. For huge sets (confirmed live: 52k `maigret`, 303k `nmap`) this OOMs the worker heap during setup, before the tool runs.

The fix direction is to push extractor target-computation into the **Query engine** so it queries results **by id + filter** instead of materializing them. Since results are fanned in by their Mongo `_id` **string** (`celery.chain_results`), the query layer must support `{'_id': {'$in': <ids>}, **filters}`.

This PR is the **verified prerequisite**: id-query support across backends, with tests. The extractor refactor itself is scoped in `docs/design/fanin-query-extractors.md` and deferred (see below).

## Per-backend findings (tested)

| backend | `{'_id':{'$in':ids}, **filters}` | |
|---|---|---|
| **api** | ✅ already works | server-side `convert_query` coerces `_id` `$in`/`$nin` strings → `ObjectId` (regression test added in secator-api#\<pending\>). Core `ApiBackend` just POSTs JSON. |
| **mongodb** | ✅ **fixed here** | backend passed the query straight to pymongo with **no** ObjectId conversion → raw-string `_id` `$in` matched **nothing**. Added `_convert_id_query` (mirrors `hooks.mongodb.get_results` + the api `convert_query`) in `_execute_search`/`_count`/`_update`. |
| **local (json)** | ✅ mechanism works; n/a by id | `match_query` supports `$in`+filters, but local findings key on `_uuid`; local never fans-in by id (addon off) — full objects are already in `context['results']`. |

`_id` is not a `PROTECTED_FIELD`, so `_merge_query` keeps the client `_id` filter while still enforcing `_context.workspace_id` + `is_false_positive` (tested).

## Changes
- `secator/query/mongodb.py`: `_convert_id_query` helper + applied in the three execute methods.
- `tests/unit/test_query.py`: `TestIdFanInQuery` + `TestMongoDBIdConversion` (mongodb ObjectId conversion incl. a regression that fails if ids reach the driver as strings; base-query merge preserves `_id`; local `$in`+filter mechanism). Mongo tests `importorskip('bson')` so CI without the mongodb extra skips cleanly.

## Deferred (design doc included)
`docs/design/fanin-query-extractors.md` scopes the extractor refactor. The hard part: extractor `condition`s are arbitrary Python `eval` (`item._source.startswith("gf")`, `len(targets)==0`, `not url.verified`) with no general Mongo translation, on the hot path of every runner. Recommended next step = push only the always-safe `_type`(+workspace) filter and field **projection** into the query, keep the Python condition on the reduced/projected set (folds in #9). Its own PR.

## Tests
`secator test unit --test test_query` → 148 passed, 8 skipped. Lint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MongoDB queries using lists of document IDs, including support for converting valid string IDs to the correct database format.
  * Ensured ID-based searches, counts, and updates preserve workspace and false-positive filtering.
  * Added support for ID fan-in queries across local JSON and MongoDB backends.

* **Documentation**
  * Added design documentation outlining improvements to reduce memory usage in fan-in query workflows.

### DIFF
--- a/docs/design/fanin-query-extractors.md
+++ b/docs/design/fanin-query-extractors.md
@@ -1,0 +1,87 @@
+# Design: query-by-id extractors (fan-in OOM, RC#6/#9)
+
+## Problem
+
+Every runner receives the **full accumulated result set** of the chain as its
+input. A `_base` runner then computes its targets from those results with
+`run_extractors` / `process_extractor` (`secator/runners/_helpers.py`). For very
+large sets this materialises the whole list on the worker Python heap during
+setup and OOMs the worker *before* the tool even runs (so `command.py`'s
+subprocess memory guard is blind to it). Two confirmed live prod zombies:
+`maigret` with 52,337 results and `nmap` with 303,499 results, each in an
+infinite OOM→requeue→OOM loop.
+
+When the MongoDB addon is enabled, results are already shrunk to their Mongo
+`_id` **string** in the chain (`celery.chain_results`) and rehydrated downstream
+by `hooks.mongodb.get_results` — but the runner still calls `get_results` to
+materialise the whole list to reason over it. The fan-in is the id list; the
+heap blow-up is the rehydration + Python-side filtering.
+
+## Goal
+
+Let a `_base` runner **query** previous results to compute its targets instead of
+holding them all — push the work into the `Query` engine
+(`secator/query/`), which filters at the datastore (cheap, especially for
+MongoDB) and returns only the matching rows.
+
+## Prerequisite (this PR): `{'_id': {'$in': ids}, **filters}` per backend
+
+Since results are fanned in by id string, the query layer must support an id +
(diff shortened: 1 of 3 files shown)


ocervell:
Closing as superseded. This id-query backend support was a prerequisite for the #1300 approach (query the fan-in by {_id:$in}). That approach is closed; the replacement #1307 filters the fan-in by _type+condition via QueryEngine and issues no id-based backend query. Verified no other caller on main uses MongoDBBackend id-queries (get_results uses db.findings.find directly). Reopen if a {_id:$in} QueryEngine caller ever appears.

PR #1299, opened 2026-07-10
feat: add live JSON driver (per-runner report.json, concurrency-safe)

ocervell:
## What

Adds `secator/hooks/json.py` — a **live JSON driver** (not an exporter). It mirrors the mongodb/sqlite drivers but persists to the local filesystem: every runner writes/updates its **own** `report.json` in `runner.reports_folder` **as results are produced** (`on_item`) and as its status changes (`on_init`/`on_start`/`on_interval`/`on_end`), rather than a single end-of-run dump.

Wired into the driver/hooks system alongside the DB drivers:
- module-level `HOOKS` dict keyed by `Scan`/`Workflow`/`Task` (same shape as `hooks/mongodb.py`, `hooks/sqlite.py`), so `_apply_context_drivers` registers it from `context['drivers']` with no special-casing;
- added to `AVAILABLE_DRIVERS` and `DRIVER_PRIORITY` in `definitions.py`.

Use it with `secator x/w/s <target> --driver json` (or `drivers.defaults: [json]`).

The emitted format (`{"info": {...}, "results": {type: [items]}}`) is byte-compatible with the JSON **exporter** and with what the **`local` query backend** (`secator/query/json.py`) already reads — the per-runner-file layout under `<reports>/<ws>/<type>s/<id>/report.json` **is** the read/merge view, so no new merge code is needed.

## Concurrency design (the crux)

Results are produced by many workers writing the report tree at once. Handled for **both** Celery pools:

- **prefork pool** → separate OS processes (separate memory)
- **gevent pool** → greenlets in one process (cooperative)

**Approach chosen: per-runner-file sharding + locked read-modify-write + atomic swap.**

1. **Sharding does most of the work.** Each runner owns its own `report.json`, so two *different* runners never contend. The existing `reports_folder` (per-runner autoincrement dir) gives this for free, and the query backend already merges these files at read time.
2. The remaining hazard is two writers on the **same** file (a redelivered task, or `update_finding` racing `update_runner`). Every write is a locked RMW:
   - **in-process `threading.Lock` per path** — gevent monkey-patches `threading`, so this is greenlet-cooperative under the gevent pool and a real mutex under prefork threads. It serializes same-process writers *before* they touch the OS lock, so the blocking `flock` below is only ever contended **across processes** (rare), never stalling the gevent hub.
   - **`fcntl.flock(LOCK_EX)` on a stable sidecar `.lock` file** — mutual exclusion across prefork **processes**. The lock is a *separate* file (never replaced), so `os.replace()` can't pull the locked inode out from under a holder (the classic flock+rename pitfall).
   - **tempfile + `os.replace`** — atomic on POSIX, so a concurrent reader (e.g. the query backend mid-run) always sees a whole old-or-new file, never a torn write.

The lock serializes RMW ⇒ **no lost updates**; the atomic swap ⇒ **no torn reads**.

### Why this over the alternatives
- **Append-only JSONL** is trivially concurrency-safe for *new* findings, but `update_finding`/`update_runner` are **in-place** updates (upsert-by-`_uuid`, status transitions). JSONL would need compaction/last-wins-on-read and would break the existing `report.json` reader. Rejected.
- **Single mutable global JSON doc** (one file for the whole run) maximizes contention and would serialize every worker on one lock. Rejected in favor of per-runner sharding.
- Per-runner files + RMW keeps the diff small, reuses the existing folder + query layout, and is correct under both pools.

### Known ceilings (marked `ponytail:` in code)
- `fcntl.flock` blocks the gevent hub **only** under genuine cross-process contention on the *same* file (redelivery) — brief and rare. The in-process lock keeps same-process greenlets off the flock entirely.
- No `on_build` PENDING-placeholder hook (mongodb/sqlite mint one so a UI tree shows not-yet-run children). Locally the query backend discovers runners by walking report dirs, so a child appears once it starts. Easy to add if a live "pending children" view is wanted.

## Tests

`tests/unit/test_json_driver.py` (9 tests, all green; lint clean):

- **Hooks**: `update_finding` insert + uuid assignment, upsert-by-uuid (no dup), non-output-type passthrough, `update_runner` info write **preserving** already-written findings, `HOOKS` structure.
- **Mid-run queryability**: the live-written `report.json` is read back by the `local` `JsonBackend` while the runner is still "running".
- **Concurrency** (the requested test):
  - `test_concurrent_prefork_processes` — N=4 OS processes append to one `report.json`.
  - `test_concurrent_gevent_greenlets` — N greenlets in a `monkey.patch_all()`'d process append to one file (isolated in a child process so patching doesn't leak into the suite).
  - `test_concurrent_mixed_processes_and_greenlets` — prefork processes **and** a gevent-greenlet process share one file (200 appends).
  - Each asserts the final file is **valid JSON**, has **exactly** the expected count (no lost updates), and **no duplicate/dropped** uuids (no corruption).

Also verified end-to-end that `context['drivers']=['json']` registers `update_finding`/`update_runner` on `on_item`/`on_end` (identical wiring to mongodb/sqlite) and that firing the lifecycle writes a live `report.json` with `info` + findings.

## Deprecation path (exporter → driver)

The JSON **exporter** is intentionally **left in place** (no behavior change for existing runs). The driver supersedes it: a driver can piggy-back **any** lifecycle event, so the end-of-run single-dump exporter becomes redundant once the driver is the default local persistence. Suggested path: ship the driver → make it the default local backend → deprecate `exporters/json.py`. Not done here to keep the diff additive and reversible.

## Downstream benefit (noted, not depended on)

Because the driver writes `report.json` **live** in the exact format the `local` query backend reads, the local query backend could eventually work **mid-run** instead of only after the final dump. This is relevant to a sibling extractor refactor — this PR neither depends on nor touches that work.

## Open questions for the owner

1. **Default enablement**: leave `json` opt-in via `--driver json`, or add it to `drivers.defaults` / gate it behind an `addons.json.enabled` flag like sqlite?
2. **`on_build` placeholders**: worth adding PENDING child docs for a live tree view locally, or is dir-walking discovery enough?
3. **Exporter deprecation timing**: OK to schedule `exporters/json.py` removal once this is the default, or keep both indefinitely?
4. **`fsync` cost**: each write `fsync`s for durability — fine for typical scan volumes; want it made configurable for very high-finding-rate runs?

## Summary by CodeRabbit

* **New Features**
  * Added a live JSON reporting driver that can persist runner metadata and findings to `report.json` during execution.
  * Enabled the JSON driver in supported driver options and adjusted driver execution priority.

* **Bug Fixes**
  * Improved JSON report reliability with atomic, lock-protected read-modify-write persistence to prevent lost updates and torn writes during concurrent runs.
  * Corrupted JSON is now reported via errors rather than silently overwritten.

* **Tests**
  * Added unit tests for the JSON driver behavior, query integration, and concurrency safety for atomic JSON persistence.

### DIFF
--- a/secator/definitions.py
+++ b/secator/definitions.py
@@ -55,7 +55,7 @@ LLM_SPINNER_MESSAGES = [
 ]
 
 # Available drivers and exporters
-AVAILABLE_DRIVERS = ['mongodb', 'gcs', 'api', 'discord', 'sqlite']
+AVAILABLE_DRIVERS = ['mongodb', 'gcs', 'api', 'discord', 'sqlite', 'json']
 AVAILABLE_EXPORTERS = ['csv', 'gdrive', 'json', 'jsonl', 'markdown', 'table', 'txt']
 
 # Canonical execution priority for drivers whose hooks share a lifecycle event.
@@ -67,7 +67,7 @@ AVAILABLE_EXPORTERS = ['csv', 'gdrive', 'json', 'jsonl', 'markdown', 'table', 't
 #      directly to the store the API reads.
 # Drivers not listed here (e.g. discord notifications, external drivers) are not
 # ranked and keep their relative order after the ranked ones.
-DRIVER_PRIORITY = ['gcs', 'mongodb', 'sqlite', 'api']
+DRIVER_PRIORITY = ['gcs', 'mongodb', 'sqlite', 'json', 'api']
 
 # Vocab
 ALIVE = 'alive'
--- a/secator/hooks/json.py
+++ b/secator/hooks/json.py
@@ -18,116 +18,25 @@
 #   Sharding does most of the work: each runner owns its own report.json, so two
 #   *different* runners never touch the same file. The remaining hazard is two
 #   writers hitting the SAME file (a redelivered task, or update_finding racing
-#   update_runner). We make every write a locked read-modify-write with an
-#   atomic swap, correct under both pools:
-#     1. per-path in-process lock (``threading.Lock``) — gevent monkey-patches
-#        threading, so this is greenlet-cooperative under the gevent pool and a
-#        real mutex under prefork threads. It serializes same-process writers
-#        BEFORE they touch the OS lock, so the blocking flock below is only ever
-#        contended across processes (rare), never stalling the gevent hub.
-#     2. ``fcntl.flock(LOCK_EX)`` on a stable sidecar ``.lock`` file — mutual
-#        exclusion across prefork processes. The lock is a *separate* file (never
-#        replaced), so os.replace() below can't pull the locked inode out from
-#        under a holder.
-#     3. tempfile + ``os.replace`` — atomic on POSIX, so a concurrent reader
-#        (e.g. the query backend mid-run) always sees a whole old-or-new file,
-#        never a torn write.
-#   The lock serializes RMW, so no update is lost; the atomic swap means no
-#   reader ever sees corruption.
+#   update_runner). Every write goes through ``atomic_json`` (secator.utils): a
+#   locked read-modify-write with an atomic swap, correct under both pools. See
+#   that function's docstring for the layered-lock details.
 #
 # ponytail: no `on_build` PENDING placeholder (mongodb/sqlite mint one so the UI
 # tree shows not-yet-run children). Locally the query backend discovers runners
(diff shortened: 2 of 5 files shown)


ocervell:
@coderabbitai review

ocervell:
Superseded by #1312 (the full stack lands there as a unit).

PR #1300, opened 2026-07-10
feat(runners): query fan-in by id+type+projection so extractors don't OOM (RC#6)

ocervell:
## ⚠️ Stacked on #1297

Base is `fix/query-id-in-backends` (**#1297**), not `main` — review/merge that first. This PR's own diff is just the extractor push-down.

## Why

RC#6 (fan-in Python-heap OOM): a child runner received the **full accumulated fan-in** and rehydrated **every** finding onto the worker heap (`get_results` in `_base.__init__` / `mark_runner_*`) *before* computing its targets. Huge sets OOM the worker during setup — confirmed live zombies: 52k `maigret`, 303k `nmap`.

This pushes extractor target-computation down to a **targeted DB fetch** so the whole set is never materialized.

## Design (driver-branched, per product-owner steer)

| driver | fan-in handling |
|---|---|
| **db (mongodb worker)** | keep the fan-in **by id** (cheap); per extractor, fetch only that `_type`'s findings with a bounded field **projection** via `get_results(ids, types, fields)`. The Python `condition` + formatting run **unchanged** on the reduced subset. |
| **local (json)** | **unchanged** in-memory `_results` path — report JSON is written only at end-of-run, so it can't be queried live. |

**Why `get_results`, not `QueryEngine.search()`:** the engine's enforced base query (`_tagged`, `is_false_positive:{$ne:True}`) is correct for a *user* query but **wrong** for rehydrating chain fan-in — during a live scan the just-produced findings are untagged, so `_tagged:True` would drop them all and change extractor targets. `get_results` is the filter-free fetch-by-id primitive. #1297's QueryEngine id-query support remains the **client/report-path** foundation.

**#9 folded in:** the fan-in flows downstream as ids (`chain_output` re-emits `prior_result_ids`); `forward_results` was already id-only for the mongodb path, so the sole materialization was `get_results` — now gone.

## Changes
- `hooks/mongodb.py`: `get_results(uuids, types=None, fields=None)` — push `_type` `$in` + include-projection into the Mongo query.
- `runners/_helpers.py`: `process_extractor` fetch-by-type branch; `_extractor_fields()` — a **bounded, over-inclusive** projection (every identifier the field/group_by/condition mention + mandatory identity/context) so it only ever omits fields the extractor never references (never under-projects → semantics identical).
- `runners/_base.py`: `_split_fanin` / `_make_fanin_fetch` / `_query_fanin_enabled`; `_run_extractors` wires the fetch. `prior_result_ids` is not in `toDict()` (no Mongo doc bloat) and only lives within one task invocation.
- `celery.py`: `chain_output()`; `mark_runner_started`/`completed` use the split.
- `docs/design/fanin-query-extractors.md`: updated to record the implementation + decisions.

## Tests — `tests/unit/test_fanin_query.py`
1. **local == db** targets for the same input/extractor (`test_run_extractors_query_matches_in_memory`).
2. db path derives targets via id+filter+projection **without** touching the full set — a `_Poison` full-set raises if iterated (`test_fetch_used_instead_of_full_set`, `test_condition_still_runs_on_fetched_subset`).
3. **300k-id fan-in** stays **< 5 MB** peak via `tracemalloc`, fetch called once, returns 1 object (`test_300k_fanin_does_not_materialize`).
4. `get_results` pushes the type filter + projection; `_split_fanin` driver-branch; projection correctness.

Full unit suite: **640 passed, 347 skipped**; lint clean. (Tests are written to survive `test_config`'s `clear_modules()` module-reload pollution.)

## Nuance (documented, acceptable)
A child no longer holds prior *finding* objects in `self.results`, so it can't incidentally dedupe its own output against the fan-in via `self.uuids` — cross-finding dedup already happens at persistence / `tag_duplicates`, and `self_results` (own-source) is unaffected.

## Before merge
Unit tests cover the logic, not the live distributed chain. Needs one **mongodb-worker integration run** (a scan with a large fan-in) to confirm end-to-end.

### DIFF
--- a/docs/design/fanin-query-extractors.md
+++ b/docs/design/fanin-query-extractors.md
@@ -74,14 +74,62 @@ path of **every** runner. A correct refactor needs either:
 
 Option 2 is the lazy, high-value first step: for `maigret`/`nmap` the `_type`
 filter alone cuts the fan-in by orders of magnitude and the projection removes
-the per-object heap cost, with **zero** change to condition semantics. It should
-be its own reviewable PR on top of this one.
+the per-object heap cost, with **zero** change to condition semantics.
+
+## Implemented — Option 2 (this stacked PR, on `fix/query-id-in-backends`)
+
+Driver-branched: **db (mongodb worker)** = fetch by id + `_type` + projection;
+**local** = unchanged in-memory `_results`.
+
+Key decision: the worker rehydrates the fan-in via `hooks.mongodb.get_results`
+(a direct, filter-free fetch-by-id), **not** `QueryEngine.search()`. The engine's
+enforced base query (`_tagged`, `is_false_positive: {$ne:True}`) is correct for a
+*user* query but **wrong** for rehydrating chain fan-in — during a live scan the
+just-produced findings are untagged, so `_tagged:True` would drop them all and
+change extractor targets. So `get_results` is the right primitive; PR #1's
+QueryEngine id-query support remains the client/report path foundation.
+
+Changes:
+- `hooks/mongodb.py::get_results(uuids, types=None, fields=None)` — push a
+  `_type` `$in` filter + include-projection into the Mongo query so an extractor
+  rehydrates only the subset (and fields) it needs.
+- `runners/_helpers.py::process_extractor` — when `ctx['fetch_by_type']` is set,
+  source this extractor's candidates from the fetch (id+type+projection) instead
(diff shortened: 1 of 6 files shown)


ocervell:
Closing — reworking this. The get_results signature change alters core fetch semantics we want to keep intact, and the Mongo-specific fan-in fetch shouldn't live in runners/_helpers. Redoing scoped to run_extractors only, with the driver-specific fetch in the MongoDB driver.

PR #1304, opened 2026-07-10
feat(query): group findings by field with aggregation

### LINKED ISSUE #1219: feat(query): group findings by field + aggregation
The commands `secator r list` and `secator r show -q`, and `secator q` should allow to group by a field, with auto-aggregation on pre-defined fields.

**Moreover:**
* `secator r list` should probably return the grouped vulnerability counts by name by default (unique vuln counts)
* `secator r show` / `secator q` should allow an argument (only for `Vulnerability` types) `--group "<field>"`  --> no need to overcomplicate the backend queries, could probably be done on processing side before displaying results (need to check if performance would suffer). It should also show (count: <COUNT>) next to the individual display lines, and `Info` messages at the end (`console.print(Info(message=f"{output_type} grouped by {field}. To show complete results, remove `--group' option`). If too many `matched_at`, truncate and display ".. and X more" to not pollute the output

**Default fields to aggregate on (if only `--group` is passed without value):**
`Vulnerability` --> `group_by: name`, auto `aggregate: matched_at`
`Tag` --> `group_by: category,name`, auto `aggregate: match`
`Technology` --> `group_by: product`, auto `aggreagte: matched_at`

==> We can add `_group_by` to Output Types on top of the existing `_sort_by` and `_table_fields` to set the default group_by fields.

When `--group` is passed on the CLI, it's okay to lose other information like `extra_data` and just use the latest finding values. If user wants detailed values, they can do so with JSON exports and other means.

ocervell:
Closes #1219.

Adds a `--group` option to `secator q` and `secator report show` that groups findings by one or more fields with auto-aggregation, done **on the processing side after the query** (no backend query changes), as the issue suggested.

## Usage

- `secator q vulnerability --group` — group vulnerabilities by their default field(s).
- `secator q vulnerability --group severity` — group by an explicit field.
- Bare `--group` uses per-type defaults; `--group "<field>[,<field>]"` overrides.

## What it does

- New `_group_by` / `_group_aggregate` class attrs on `OutputType`, set for the three types the issue lists:
  - `Vulnerability` → group by `name`, aggregate `matched_at`
  - `Tag` → group by `category,name`, aggregate `match`
  - `Technology` → group by `product`, aggregate `match`
- `group_findings()` collapses each group to the **newest** finding, keeping a `_group_count` and a **truncated** list of the aggregated field's values (`.. and X more` past 5).
- Console output shows `(count: N)` next to each grouped line.
- An `Info` message per grouped type is printed at the end (`<type> grouped by <field>. To show complete results, remove the --group option.`).

## Degradation

- Result types without `_group_by` (and non-finding results) pass through unchanged.
- `--group <field>` with no groupable finding types present → warning, no silent mis-answer.
- `--group` combined with `--format` is ignored with a warning (format already reshapes output).

## Notes / assumptions

- Grouping happens post-query for **all** backends (local/mongodb/api/sqlite) since it operates on the returned findings — no `$group` pipeline needed, matching the issue's "processing side before displaying" guidance.
- The issue's "aggregate: matched_at" for `Technology` is a typo — `Technology` has no `matched_at` field; its location field is `match`, so that is used (same as `Tag`).
- Scope: implemented the shared `--group` path for `secator q` / `report show`. The issue's hedged "`secator r list` should probably default to grouped vuln counts" is a separate command/code path and was left out to keep scope tight — easy follow-up if wanted.

## Test

Added `TestGroupFindings` in `tests/unit/test_query_utils.py` (group-by-name + aggregate + truncation asserts). `secator test unit` query/report/output-type suites pass (240 tests).

## Summary by CodeRabbit

* **New Features**
  * Added an optional `--group` flag to query and report display commands.
  * Group results by selected fields or recommended defaults for supported finding types.
  * Display grouped result counts and summarized aggregated values.
  * Show guidance when grouping is unavailable or used with formatted output.

* **Tests**
  * Added coverage for grouping, aggregation, and truncated value displays.


ocervell:
@claude address coderabbit review comment

PR #1305, opened 2026-07-10
feat(security): run raw-socket tools via caps not sudo (drop GTFObins nmap root)

ocervell:
## Problem

The prod worker Docker image (`.docker/Dockerfile.alpine`) grants the `secator`
user `NOPASSWD: /usr/bin/nmap *` (+ `arp-scan`, `arp`) in `/etc/sudoers`. That's
a classic [GTFObins](https://gtfobins.github.io/gtfobins/nmap/) root shell —
`sudo nmap --script=/tmp/x.nse <target>` executes arbitrary NSE Lua as root —
so any shell-injection vulnerability reached inside a worker process escalates
straight to root.

## Fix

Run the raw-socket tools (nmap, arp-scan, arp, naabu) via **Linux file
capabilities** (`CAP_NET_RAW`/`CAP_NET_ADMIN`) instead of `sudo`, gated by a new
config flag so local/dev installs keep the simpler sudo path:

- `secator/config.py`: `Security.privileged_mode: str = 'sudo'` — `'sudo'`
  (default, prepends `sudo`) or `'caps'` (rely on setcap'd binaries). Overridable
  via `SECATOR_SECURITY_PRIVILEGED_MODE`.
- `secator/runners/command.py`: the `requires_sudo` sudo-prepend (`self.cmd`/
  `self.cmd_redacted`) is now gated by `CONFIG.security.privileged_mode != 'caps'`.
- `.docker/Dockerfile.alpine` (flavor-aware):
  - **full** image (prod): no `NOPASSWD` sudoers grant for nmap/arp-scan/arp.
    After tools are baked in, `setcap cap_net_raw,cap_net_admin+eip` is applied
    to the installed binaries (`/usr/bin/nmap`, `/usr/bin/arp-scan`, `/sbin/arp`,
    `/home/secator/.local/bin/naabu`), and `security.privileged_mode` is set to
    `caps` for the `secator` user.
  - **lite** image: unchanged — tools install at runtime so there's nothing baked
    to `setcap`, so it keeps the old `NOPASSWD` sudoers grants and default
    sudo mode. This is a known deploy-time residual; prod should prefer the
    full image.

## Test plan

- [x] Added `tests/unit/test_command.py::TestCommandPrivilegedMode` — with
      `privileged_mode = 'caps'` a `requires_sudo` command's `self.cmd` does NOT
      get `sudo ` prepended; with `'sudo'` (default) it does. `2 passed`.
- [x] `flake8` clean on the changed Python files.
- [ ] Dockerfile change is CI-validated only (tool install needs network access
      not available in this environment) — full build happens on this PR's CI.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01P5vSjfkBuGAAHdKxHS3ySm

### DIFF
--- a/.docker/Dockerfile.alpine
+++ b/.docker/Dockerfile.alpine
@@ -41,10 +41,18 @@ RUN apk add --no-cache \
 
 RUN adduser -D secator
 
-# Allow runing 'nmap' with sudo; for priviledged scans
-RUN echo "secator ALL=(ALL) NOPASSWD: /usr/bin/nmap *" >> /etc/sudoers && \
-    echo "secator ALL=(ALL) NOPASSWD: /usr/bin/arp-scan *" >> /etc/sudoers && \
-    echo "secator ALL=(ALL) NOPASSWD: /sbin/arp *" >> /etc/sudoers
+# Raw-socket tools (nmap/arp-scan/arp) need CAP_NET_RAW/CAP_NET_ADMIN.
+# full flavor: no baked-in sudoers grant here -- tools are installed below, then
+# switched to file capabilities (see the setcap block after tool install). A
+# `NOPASSWD: /usr/bin/nmap *` sudoers entry is a GTFObins root shell (a shell
+# injection in a worker can `sudo nmap --script=/tmp/x.nse ...` -> root).
+# lite flavor: tools are installed at runtime, so there are no baked binaries to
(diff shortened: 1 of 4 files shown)


ocervell:
This will cause some issues with tools that are only installed as root (e.g: arp), need to be refined

PR #1321, opened 2026-07-30
feat(output-types): tag known-exploited vulnerabilities with `kev` (#1319)

### LINKED ISSUE #1319: feat: add KEV tag to any vulns emitted by Secator
* At startup, add to the files downloaded (along with the wordlists etc...) the KEV JSON (https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json) and extract a variable `KEV_CVE_IDS  = [vuln['cveId'] for vuln in KEV_DATA['vulnerabilities']]`
* In the vulnerability.py output types, when initializing a vuln, if the vuln id is set, add the `kev` tag if it's in `KEV_CVE_IDS`.

juandresrodca:
Closes #1319.

Adds kev tagging for known-exploited vulnerabilities, following the recipe in the issue.

## notes

- **Lazy + memoized**, not fetched at import time, so offline mode and tests don't hit
  the network. Fails open: offline / download error / malformed feed → empty set, so
  tagging is always a safe no-op.
- Matches `self.id` against the CVE list per the recipe; matching CVEs found in
  `references` / `extra_data` could be a follow-up.
- Happy to wire the fetch into startup explicitly instead of lazy-on-first-use, or to
  expose the URL / a toggle in `Config`, if you'd prefer.

---
- **secator/kev.py (new):** downloads the CISA KEV feed once
  (https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json),
  cached to the data dir via the existing download_file() helper (same path as
  wordlists/payloads), and exposes a memoized get_kev_cve_ids() returning the set
  of KEV CVE ids — built as {v['cveId'] for v in KEV_DATA['vulnerabilities']}.
- **secator/output_types/vulnerability.py:** __post_init__ tags a vuln with kev
  when its id (the CVE id) is in the KEV set. Done here so it applies to vulns from
  **every** task, not just CVE-enriched ones — which lights up the already-documented
  vulnerability.tags ~= 'kev' cheatsheet query.
- **tests/unit/test_kev.py (new):** unit tests (no network — the loader is patched).

## Testing
secator test unit --test test_kev (6 passed); flake8 clean. Existing
test_output_types still green.

## Summary by CodeRabbit

- **New Features**
  - Vulnerabilities associated with CISA Known Exploited Vulnerabilities are now automatically marked with a `kev` tag.
  - CVE matching is case-insensitive, and duplicate tags are avoided.
  - The latest KEV catalog is downloaded and cached for efficient reuse.

- **Bug Fixes**
  - Vulnerability processing now continues safely when the catalog is unavailable, malformed, or a vulnerability has no identifier.

### DIFF
--- a/secator/kev.py
+++ b/secator/kev.py
@@ -1,0 +1,55 @@
+"""CISA Known Exploited Vulnerabilities (KEV) catalog helpers.
+
+The CISA KEV catalog (https://www.cisa.gov/known-exploited-vulnerabilities-catalog)
+lists CVEs that are known to be actively exploited in the wild. Secator downloads the
+feed once (cached to the data directory, like wordlists / payloads) and exposes the set
+of KEV CVE IDs so that any emitted vulnerability whose CVE is known-exploited can be
+tagged with ``kev``.
+"""
+import json
+
+from secator.config import CONFIG, download_file
+
+KEV_URL = 'https://www.cisa.gov/sites/default/files/feeds/known_exploited_vulnerabilities.json'
+
+# Lazily-loaded cache of upper-cased KEV CVE IDs. ``None`` means "not loaded yet"; an empty
+# set means "loaded but unavailable" (offline / download failed) so we never retry per-vuln.
+_KEV_CVE_IDS = None
+
+
+def get_kev_cve_ids():
+	"""Return the set of KEV CVE IDs (upper-cased), downloading the feed once if needed.
+
+	The result is memoized for the lifetime of the process. On any failure (offline mode,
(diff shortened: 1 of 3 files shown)


juandresrodca:
Let me know if you need any explain on the code or modify it

ocervell:
Hi @juandresrodca ! Sorry I just returned from vacations. 
This is really great ! I wanted to do it when coming back actually.
Give me a little time to review and it will go into the next release, thanks a lot for taking the time to make a PR ;)

PR #1331, opened 2026-08-18
feat(kev): bundle a local CISA KEV mirror as an offline fallback

ocervell:
## Why
KEV tagging (#1321) downloads the CISA feed from `cisa.gov` at run time. In offline / constrained environments that host isn't reachable, so `kev` tagging silently does nothing. This bundles the catalog with the package and reads it from a local cache — **no network at run time**.

## Behavior
- **Bundled mirror** — `secator/data/known_exploited_vulnerabilities.json` ships with the package (1666 CVEs, minified ~1.2MB). Shipped via a `[tool.hatch.build] artifacts` entry (needed because `*.json` is git-ignored → hatchling would otherwise exclude it; **verified present in the built wheel**).
- **On first use (startup)** the bundle is copied into the cache dir (`CONFIG.dirs.cves`) — re-seeded only when the cache is missing or older than the bundle (e.g. after a `pip` upgrade ships a newer catalog).
- **Reads come from the cache** — `get_kev_cve_ids()` loads once from the cached JSON, no per-run download. KEV tagging works offline out of the box.
- **`secator update`** calls `refresh_kev()`: downloads the catalog from cisa.gov, **updates the bundled mirror and re-copies it to the cache**. Fail-safe — keeps existing files on offline mode or any download error.
- **`.github/workflows/update-kev.yml`** — nightly (`workflow_dispatch` too) refresh of the bundled mirror via PR (matching `update_tools_version.yml`), only when it changed.

## Also fixes a pre-existing bug
The parser used `vuln['cveId']` but the CISA feed key is **`cveID`** (capital I/D). Guarded by `.get('cveId')`, it skipped **every** entry — so KEV tagging was a permanent no-op *even online*. It went unnoticed because the unit tests mock the loader. Verified fixed: loading the real bundle tags `CVE-2021-44228` (Log4Shell).

## Tests
`tests/unit/test_kev.py`: parse (uppercasing + skip-missing `cveID`), cache seeding from the bundle, cache-first read (no network), refresh writes both bundle + cache, offline no-op — isolated from the committed mirror so `refresh()` can't touch it. **11 passed, flake8 clean.**

## Summary by CodeRabbit

* **New Features**
  * Added automatic updates for the CISA Known Exploited Vulnerabilities catalog.
  * Included an offline catalog fallback so vulnerability data remains available without network access.
  * The update command now refreshes vulnerability data and reports the number of loaded CVEs.
  * Packaged the vulnerability catalog for reliable distribution.

* **Bug Fixes**
  * Improved handling of missing, outdated, invalid, or unavailable catalog data.


ocervell:
@coderabbit review the PR again

PR #1343, opened 2026-08-21
fix(security): enforce generic target-scope allow-list on discovered targets

ocervell:
## Problem

Domain scans discover subdomains via `subdomain_recon` and fan them into `host_recon` and the `url_*` workflows (`configs/scans/domain.yaml`: `host_recon.targets_: [{type: subdomain, field: host}]`). Those **dynamically-discovered** targets were never checked against any scope, so a run authorized for an apex only (e.g. `claris.com`, not `*.claris.com`) would still scan every discovered subdomain — a **scope escape**.

Core has no concept of the platform's mandates and must stay that way, so this adds a *generic* scope filter that the platform (secator-api) feeds.

## Fix

- **`secator/scope.py`** — neutral host allow/deny matcher (`host_in_scope`, `target_in_scope`). Rules (regex / CIDR / `*.` wildcard / exact host) mirror the platform matcher so a target the ingress gate would reject is dropped identically mid-scan. Deny wins; empty `in_scope` = allow-all (unchanged behavior).
- **`run_extractors`** (`secator/runners/_helpers.py`) — the single fan-in choke point every runner's inputs flow through. Filter resolved `inputs` (user-supplied **and** discovered) against `in_scope`/`out_of_scope` run-options; dropped targets logged under `debug(sub='scope')`. Run-options propagate scan→workflow→task verbatim, so `host_recon` gets the scope for free.

The platform side (deriving `in_scope`/`out_of_scope` from the mandate and injecting them) is a companion PR in **secator-api**.

## Tests

`tests/unit/test_scope.py` — matcher semantics (apex vs wildcard, CIDR, regex, deny-wins) + `run_extractors` dropping a discovered out-of-scope subdomain. All pass; no regression in extractor/helper tests.

## Summary by CodeRabbit

* **New Features**
  * Added flexible target-scope matching for hosts, subdomains, IP addresses, CIDR ranges, URLs, and regular expressions.
  * Added allowlists and denylists for controlling which resolved or discovered targets are processed.
  * Deny rules take precedence over allow rules, with safe handling of invalid scope values.

* **Bug Fixes**
  * Prevented out-of-scope inputs from being passed to extractors.

### DIFF
--- a/secator/runners/_helpers.py
+++ b/secator/runners/_helpers.py
@@ -4,6 +4,7 @@ import re
 from secator.config import CONFIG
 from secator.output_types import Error
 from secator.query.ast import substitute_ctx_constants
+from secator.scope import as_scope_list, host_in_scope
 from secator.utils import deduplicate, debug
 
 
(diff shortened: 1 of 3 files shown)


ocervell:
Closing as **superseded**, not merging — merging this branch would regress `main`.

This branch (`claude/scope-escape-fix`) is the original, simpler version of the discovered-target scope-escape fix. Its content shipped to `main` in its **evolved** form via the scope-gate promotion **#1350** (released as core **v0.43.0**, now live on prod), together with the follow-on work that built on it:

- #1345 single-source target classifier (`classify_target` / `canonicalize_target`)
- #1346 secure scope matcher (anchored-fullmatch regex, IPv6 containment, deny-wins)
- #1347 `[IPv6]:port` canonicalization fix
- #1348 scope-inheritance-to-discovered-targets tests + producer gap
- #1349 `[WRN]` on out-of-scope drop

Diffing this branch against current `main` is **net −787 lines** (it would delete the classifier, `test_scope_inheritance.py`, `test_utils.py`, and revert `scope.py` to the earlier form). The `run_extractors` `host_in_scope` enforcement this PR introduced is present on `main` and live on prod. Closing.

PR #1351, opened 2026-08-25
test(integration): stop pinning the drifting free.fr testssl cert fingerprint

ocervell:
`test_tasks` hard-coded a Certificate fingerprint for the live host **free.fr** (`outputs.py:164`, `B425AF159E…`). free.fr renewed its cert (now `0F098362CA…`, valid to 2026-11-01), so the fixture stopped matching and the `integration` gate went red — the same red check that got bypassed during the 0.43.0 promotion.

Fix (mirrors #1325's drift handling): drop the exact-match fixture and verify the cert via `OUTPUTS_CHECKS` — assert testssl produced a Certificate with a real 64-char sha256 fingerprint and a host, without pinning the value that renews. `empty_results_allowed=False` still ensures testssl produces output.

Test-only change; verified the check passes a real cert and fails an empty one.

### DIFF
--- a/tests/integration/outputs.py
+++ b/tests/integration/outputs.py
@@ -18,7 +18,19 @@ OUTPUTS_CHECKS = {
 				},
 			],
 			'runner': '^(?!urlfinder|xurlfind3r|gau|urlparser$).*',
-		}
+		},
+		Certificate: {
+			'checks': [
+				{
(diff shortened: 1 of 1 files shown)


ocervell:
Redundant — the fix only needs to unblock main's release gate; canary picks it up on its next reset-to-main. Superseded by #1352 (to main).

PR #1353, opened 2026-08-25
chore: cut 0.43.1 (green integration suite)

ocervell:
Triggers a clean bugfix release **0.43.1** via `Release-As`, now that the integration gate is green (testssl cert-fixture drift fixed in #1352). `test:` commits don't bump the version, so this forces the patch. No runtime change vs 0.43.0.

Will wait for all checks green before merging (golden rule).


ocervell:
Closing — wrong mechanism for forcing a release. The right way is an empty semantic commit (e.g. `fix: …` via `git commit --allow-empty`) that release-please recognizes, which opens/updates the release PR.

PR #1357, opened 2026-08-25
fix(wpprobe): restore -f multi-target scanning

Chocapikk:
Follow-up to #1355 by @bpce-it-dis-soc-sof, which worked around Chocapikk/wpprobe#33 by scanning one URL at a time (`input_chunk_size = 1`, `-f` disabled), because wpprobe emitted concatenated JSON objects (JSONL) that `json.load` could not parse for multiple targets.

That's now fixed upstream in **wpprobe v0.12.11**, which always emits a proper JSON array, one object per target (Chocapikk/wpprobe#33). So `-f` can be restored and targets scanned in a single wpprobe call again.

### Changes
- Re-enable `file_flag = '-f'` and drop `input_chunk_size = 1` (from #1355)
- Bump `install_version` to `v0.12.11`
- Parse the JSON array (one object per target), still accepting the old single-object shape for backward compatibility

### Testing
Ran against a live WordPress lab (Elementor 3.5.0 installed) with two targets:

```
printf 'http://localhost:8099\nhttp://127.0.0.1:8099\n' | secator x wpprobe -json
# -> 4 tags, 64 vulnerabilities, split across both URLs, correct CVEs
```

Single-target runs are unchanged (2 tags / 32 vulns), and `flake8` is clean.

Refs:
- freelabz/secator#1355
- Chocapikk/wpprobe#33

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated WordPress scanning to support the latest wpprobe output format.
  * Improved processing of multiple targets, including their URLs, plugins, and themes.
  * Maintained compatibility with legacy wpprobe result formats.
  * Updated the bundled wpprobe version for improved scanning reliability.

### DIFF
--- a/secator/tasks/wpprobe.py
+++ b/secator/tasks/wpprobe.py
@@ -18,10 +18,9 @@ class wpprobe(Command):
 	"""Fast wordpress plugin enumeration tool."""
 	cmd = 'wpprobe'
 	input_types = [URL]
-	input_chunk_size = 1
 	output_types = [Vulnerability, Tag]
 	tags = ['vuln', 'scan', 'wordpress']
-	# file_flag = '-f'
+	file_flag = '-f'
 	input_flag = '-u'
 	opt_prefix = '-'
 	opts = {
@@ -34,7 +33,7 @@ class wpprobe(Command):
 	opt_key_map = {
 		THREADS: 't'
 	}
-	install_version = 'v0.11.1'
+	install_version = 'v0.12.11'
 	install_cmd = 'go install github.com/Chocapikk/wpprobe@[install_version]'
 	github_handle = 'Chocapikk/wpprobe'
 	install_post = {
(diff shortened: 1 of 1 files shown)


Chocapikk:
Thanks @coderabbitai. This PR intentionally keeps the change minimal: the wpprobe task already does all of its parsing in `on_cmd_done` (`_parse_software`, `_parse_vulns`, `_normalize_severities`), and #1355 kept it there too. My change follows that existing pattern rather than introducing a new one.

There's also a practical reason `on_cmd_done` fits here: wpprobe writes a single complete JSON document (an array of per-target objects) to a file, so it has to be loaded whole and iterated, rather than streamed one item per line the way `item_loaders` is designed for.

Migrating the whole wpprobe parser to `item_loaders` is a reasonable cleanup, but it's a larger refactor of pre-existing code and out of scope for a regression fix. Happy to do it as a follow-up PR if a maintainer wants it. @ocervell what's your preference?

ocervell:
Thanks @Chocapikk for the fast move on this ! Greatly appreciated, will test and merge this shortly 😉

PR #1362, opened 2026-08-30
fix(ai): accurate + resilient empty-response handling

ocervell:
## Report (canary)
Using `openrouter/anthropic/claude-opus-4.8` with the AI task failed with:
> 3 consecutive empty responses - the model may not support tool calling. Stopping.

## Investigation
Pulled the failing runner from canary mongo → model was `openrouter/anthropic/claude-opus-4.8`. Reproduced the raw `litellm.completion` call from inside the canary pod:
- **Trivial tool:** `finish_reason: tool_calls`, content **and** a tool call returned ✅
- **Real chat-mode tool set (5 tools) + full 14KB system prompt:** `finish_reason: tool_calls`, content + tool call ✅

So the model **does** support tool calling, and `call_llm` parses it correctly. The failure was empty responses from the provider (not reproducible now → a transient/upstream condition), which the loop **misdiagnosed**.

## Root cause
The agent loop treats *any* turn with no content and no tool_calls as "the model may not support tool calling" and stops after **3 immediate** retries. But an empty turn is almost always:
- a **transient provider hiccup** (empty upstream response), or
- a **truncated** turn (`finish_reason == "length"`, hit the output-token cap).

A wrong model name 400s rather than returning empty, so "doesn't support tools" is essentially never the real cause.

## Fix
- `call_llm` returns `finish_reason`.
- The empty-response path reports the **real** cause (truncation vs empty-from-provider, including `finish_reason`) instead of the misleading tool-calling claim.
- Empty-retry now **backs off** (2^n, capped 8s) so a transient blip can recover, instead of 3 immediate re-calls into the same blip.

## Tests
`tests/unit/test_ai_resilience.py`: error no longer blames tool calling; `finish_reason=length` reports truncation; `time.sleep` patched instant. 30 passed (103 across the AI suite).

Base is `canary` because the AI stack lives there (far ahead of main).

### DIFF
--- a/secator/ai/utils.py
+++ b/secator/ai/utils.py
@@ -752,8 +752,12 @@ def call_llm(
 
 	# Get tool calls
 	tool_calls = getattr(message, 'tool_calls', None) or []
+	# finish_reason distinguishes a genuinely-empty turn from a truncated one
+	# (finish_reason == "length") — the caller uses it to report the real cause
+	# of an empty response instead of blaming tool-calling support.
+	finish_reason = getattr(response.choices[0], 'finish_reason', None)
 
-	return {"content": content, "usage": usage, "tool_calls": tool_calls}
+	return {"content": content, "usage": usage, "tool_calls": tool_calls, "finish_reason": finish_reason}
 
 
 MODEL_COLORS = [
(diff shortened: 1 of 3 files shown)


ocervell:
Folded into the unified AI PR #1288 (feat/ai-unified) instead of a standalone PR; also on canary for testing.

PR #1369, opened 2026-09-02
fix(bbot): add python dev headers to apt/dnf install prereqs

ocervell:
## Bug
The `platform (ubuntu)` CI build fails installing bbot:
```
[INF] 🔧 Installing bbot ...
⚡ pipx install bbot==2.7.2 --force
    regex_3/_regex.c:50:10: fatal error: Python.h: No such file or directory
    yara-python.c:20:10:    fatal error: Python.h: No such file or directory
[ERR] Failed to install bbot: InstallerStatus.INSTALL_FAILED
```

## Root cause
bbot's native deps (`regex`, `yara-python`) have **no prebuilt wheels for Python 3.14** (the CI Docker image's Python), so pip builds them **from source** — which needs the **Python dev headers** (`Python.h`) plus a C toolchain.

`bbot.install_pre` only carried `python3-dev` under the **`apk`** (Alpine) key. On Ubuntu/Debian (`apt`) the resolver falls through to `*` — which had `gcc`, `git`, … but **no python3-dev** — so the source build failed. (On older Pythons this was masked because prebuilt wheels existed and no source build happened.)

## Fix
Add explicit prereq entries with the Python dev headers:
- `apt` → `python3-dev` (covers ubuntu/debian/kali)
- `dnf|yum|zypper` → `python3-devel` (fedora/centos/rhel/opensuse)

Also glob the python minor version in the `install_post` ansible-collections cleanup (`python3.12` → `python3.*`) so it isn't silently skipped on 3.14+.

## Verification
Unit test asserts the apt/dnf paths resolve to the headers (mirrors `PackageInstaller`'s first-match logic); loader suite green. Not transient — a genuine missing build dependency exposed by the Python 3.14 bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BBOT installation compatibility on apt-, dnf-, yum-, and zypper-based systems by installing the required Python development dependencies.
  * Updated post-install cleanup to support ansible collections across Python 3 minor versions, rather than targeting only Python 3.12.

* **Tests**
  * Added coverage verifying package-manager-specific dependencies and version-agnostic cleanup behavior.

### DIFF
--- a/secator/tasks/bbot.py
+++ b/secator/tasks/bbot.py
@@ -270,14 +270,23 @@ class bbot(Command):
 			'extra_data': 'extra_data',
 		}
 	}
+	# bbot's deps (regex, yara-python) ship no prebuilt wheels for newer Pythons
+	# (e.g. 3.14), so pip builds them from source — which needs the Python dev
+	# headers (Python.h) plus a C toolchain. The apt/dnf paths lacked
+	# python3-dev/python3-devel, so `secator install bbot` failed on Ubuntu/Debian
+	# with `fatal error: Python.h: No such file or directory`.
 	install_pre = {
 		'apk': ['python3-dev', 'linux-headers', 'musl-dev', 'gcc', 'git', 'openssl', 'unzip', 'tar', 'chromium'],
+		'apt': ['python3-dev', 'gcc', 'git', 'openssl', 'unzip', 'tar', 'chromium'],
+		'dnf|yum|zypper': ['python3-devel', 'gcc', 'git', 'openssl', 'unzip', 'tar', 'chromium'],
 		'*': ['gcc', 'git', 'openssl', 'unzip', 'tar', 'chromium']
 	}
 	install_version = '2.7.2'
 	install_cmd = 'pipx install bbot==[install_version] --force'
 	install_post = {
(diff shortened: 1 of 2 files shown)


ocervell:
### Root cause: rolling base Python, not a dependency change

This regressed with **no code change** because `.docker/Dockerfile.ubuntu` uses `FROM ubuntu:latest` (rolling), and Ubuntu's default `python3` rolled **3.12 → 3.14**. C extensions always need `Python.h` to build from source — they were just never built before, because prebuilt wheels matched the Python. On 3.14 there are none:

| dep | bbot 2.7.2 pin | wheels |
|---|---|---|
| yara-python | `>=4.5.1,<5.0.0` | newest (4.5.4) tops out at **cp313** — no cp314 wheel exists → source build |
| regex | `>=2024.4.16,<2025.0.0` | `regex 2026.x` ships cp314 wheels, but bbot's `<2025.0.0` cap excludes it → stuck on 2024.x (no cp314) → source build |

On 3.12 both installed as cp312 wheels (no compiler/headers). On 3.14 both fall back to source and need `Python.h`, which `Dockerfile.ubuntu` never installed.

Keeping `ubuntu:latest` rolling is intentional (catch platform-edge issues early); this fix is the forward-compatible complement — source builds now work on whatever Python Ubuntu ships next, rather than requiring a matching prebuilt wheel.

Issue #1366, opened 2026-09-02
fix: wpscan crashing on exegol


bpce-it-dis-soc-sof:
On exegol, the previously installed wpscan seems to conflict with secator-installed one:
```
secator x wpscan http://localhost:8889 --sync --verbose
⚡ wpscan --force --verbose -o /root/.secator/reports/default/tasks/58/.outputs/wpscan.json --url http://localhost:8889 -f json --max-threads 50
/usr/local/rvm/rubies/ruby-3.2.2/lib/ruby/3.2.0/rubygems.rb:263:in `find_spec_for_exe': can't find gem wpscan (>= 0.a) with executable wpscan (Gem::GemNotFoundException)
	from /usr/local/rvm/rubies/ruby-3.2.2/lib/ruby/3.2.0/rubygems.rb:282:in `activate_bin_path'
	from /root/.local/bin/wpscan:25:in `<main>'
	from /root/.local/bin/ruby_executable_hooks:22:in `eval'
	from /root/.local/bin/ruby_executable_hooks:22:in `<main>'
[ERR] wpscan Could not find JSON results in /root/.secator/reports/default/tasks/58/.outputs/wpscan.json
[ERR] wpscan Command failed with return code 1
   Traceback (from command output):
   /usr/local/rvm/rubies/ruby-3.2.2/lib/ruby/3.2.0/rubygems.rb:263:in `find_spec_for_exe': can't find gem wpscan (>= 0.a) with executable wpscan (Gem::GemNotFoundException)
        from /usr/local/rvm/rubies/ruby-3.2.2/lib/ruby/3.2.0/rubygems.rb:282:in `activate_bin_path'
        from /root/.local/bin/wpscan:25:in `<main>'
        from /root/.local/bin/ruby_executable_hooks:22:in `eval'
        from /root/.local/bin/ruby_executable_hooks:22:in `<main>'
   \[[1;31mERR[0m\\] [2mwpscan[0m Could not find JSON results in /root/.secator/reports/default/tasks/58/.outputs/wpscan.json
[INF] JSON report written to /root/.secator/reports/default/tasks/58/report.json
[INF] Task wpscan finished with status SUCCESS and found 0 findings
[INF] Exporting results with exporters: csv, txt, markdown
[INF] CSV report written to /root/.secator/reports/default/tasks/58/report_target.csv
[INF] Saved TXT reports to /root/.secator/reports/default/tasks/58/report_target.txt
[...]

PR #1373, opened 2026-09-04
fix(command): resolve tools exposed as shell aliases before auto-installing

ocervell:
## Motivation
Groundwork for the Exegol integration (and a general robustness fix). Some environments expose a tool **only as a shell alias**, not a real PATH binary — e.g. Exegol's `wpscan` is an RVM-gemset alias. secator's presence check runs in a **non-interactive subprocess** where aliases aren't loaded, so it saw the tool as *missing* and auto-installed its own copy, which shadowed the working one (the root of the #1366 class of problems, and unnecessary duplication of tools the host already ships).

## Change
- **`is_installed()` → `shutil.which`**: pure-Python, cross-platform (works on Windows), no subprocess.
- **Alias fallback before auto-install**: when a tool isn't on PATH, resolve it via the login shell's `alias` builtin (`$SHELL -ic 'alias <name>'`). If it's an alias, splice the expansion into the command (preserving any `sudo`/`proxychains` prefix) so the subprocess runs the **host's** tool, and skip the install. Only auto-install when the tool is genuinely absent.

### Why `alias`, not `type`
`type`'s output is **locale-translated** — a French-locale bash prints `mytool est un alias vers « … »`, which no English regex matches. The `alias` builtin prints the literal, locale-independent `[alias ]name='<expansion>'` in both bash and zsh (and we force `LC_ALL=C` for good measure). Verified against real bash/zsh output.

## Behavior matrix
| tool state | before | after |
|---|---|---|
| on PATH | used | used (via `shutil.which`) |
| alias only (Exegol/RVM) | **auto-installed a shadow copy** | **resolved + host tool used**, no install |
| genuinely missing | auto-installed | auto-installed (unchanged) |

## Tests
- `tests/unit/test_shell_alias_resolution.py` — bash/zsh alias formats, non-alias, no-shell, shell-error, splice preserves sudo prefix + word-boundary, is_installed is PATH-only.
- `mock_command` updated to mark mocked commands installed (they never run/install; is_installed no longer routes through the mocked `subprocess.Popen`).
- Ran all mock_command-based unit suites (runners/tasks/celery/on_build): **100 passed**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Commands configured as shell aliases can now be resolved and executed when they are not available on the system PATH.
  - Alias resolution supports Bash and Zsh formats, including commands prefixed with `sudo`.
  - Improved command availability checks prevent unnecessary installation attempts when tools are already accessible through PATH.
  - Command execution now handles missing shell configuration and shell errors more reliably.

### DIFF
--- a/secator/runners/command.py
+++ b/secator/runners/command.py
@@ -523,13 +523,19 @@ class Command(Runner):
 					yield Error(message=error)
 					return
 
-			# Prepare cmds
-			command = self.cmd if self.shell else shlex.split(self.cmd)
-
-			# Check command is installed and auto-install
+			# Resolve the command before exec: prefer a PATH binary; if none, try a
+			# shell alias. Some environments (Exegol/RVM) expose tools ONLY as shell
+			# aliases, which live in the shell runtime — not on PATH — so which /
+			# shutil.which miss them. When a tool resolves only via an alias, splice the
+			# expansion into the cmd so the subprocess can run it, and skip installing a
+			# conflicting copy. Auto-install only when the tool is nowhere to be found.
 			if not self.no_process and not self.is_installed():
-				self.debug('command is not installed, auto-installing', sub='start')
-				if CONFIG.security.auto_install_commands:
+				alias = self._resolve_shell_alias(self.cmd_name)
+				if alias:
+					self._splice_resolved_command(alias)
+					self.debug(f'{self.cmd_name} not on PATH; using shell alias -> {alias}', sub='start')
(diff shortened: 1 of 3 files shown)


ocervell:
This will spawn one subprocess per command ran. On large installs where we have a Secator worker with 100s of concurrent gevent processes, this might cause some kind of process creation bomb, so maybe it needs to be revised with a lighter approach (can we check output of shutil.which and then resolve the symlink and use that ?)

PR #1392, opened 2026-09-15
fix(nmap): keep all IDS mass-scan ports at low confidence

ocervell:
## Problem
On a host flagged as an IDS mass-scan (`>20` open ports), the nmap parser demotes `global_confidence` to `low` and tags ports `ids`. But the per-port loop did `if service_confidence != 'low': global_confidence = 'high'` on a **shared, never-reset** variable — so the first port with a confident service banner flipped the whole host back to `high` for every subsequent port.

On a real OVH IDS host this left **~64,700 of ~128,854** noise ports mislabeled `confidence='high'` (nmap reports them `tcpwrapped`, `conf=8` → `service_confidence='high'`). It's why filtering IDS noise on `confidence` alone doesn't work — the `ids` tag is the reliable marker.

## Fix
Gate the upgrade with `not is_mass_scan`, so an IDS host's ports all stay `low`. `service_confidence` is still recorded per-port, so a genuine service isn't lost.

## Test
`TestNmapIdsConfidence` — a 25-port host with one confident service asserts every port stays `confidence='low'` with the `ids` tag (and the confident port keeps `service_confidence='high'`). Without the fix, ports 2–25 would be `high`.

Prod backfill of the ~64,700 already-mislabeled docs is a separate follow-up. Investigation INV-03.

https://claude.ai/code/session_01RStBpZsZ8QtJYRbbKvd4MM

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved confidence handling for hosts identified as mass scans. High-confidence service banners no longer incorrectly raise the host’s global confidence.
  - Preserved low confidence and IDS classification for mass-scan hosts while retaining accurate per-port service confidence.

- **Tests**
  - Added coverage validating confidence and tagging behavior for hosts with many open ports.


secator/tasks/nmap.py
```diff
@@ -248,7 +248,15 @@ def __iter__(self):
...
+				# `tcpwrapped` = a port that completes the TCP handshake then drops the
+				# connection with no service data — a confirmed non-service, the
+				# signature of an IDS/firewall answering probes on every port. nmap
+				# still scores it conf=8 ('high'), which used to flip the shared
+				# global_confidence to 'high' and mislabel ~half an IDS host's ports.
+				# Scope the demotion to tcpwrapped only (see the per-port `confidence`
+				# below): it must neither upgrade the host confidence nor be reported
+				# high itself, while genuinely fingerprinted services still can.
+				if service_name != 'tcpwrapped' and service_confidence != 'low':
 					global_confidence = 'high'
```
ocervell: in a context where there are 10 ports and there is a masscan detected:

nfs --> conf = 8
tcpwrapped --> conf = 8
ftp --> conf = 8
tcpwrapped --> conf = 8
rdp --> conf = 1
...

nfs will have confidence 'high' and service_confidence 'high' and switch the global confidence to 'high'
tcpwrapped will have confidence 'high' (because it was switched by the prior port) and service confidence 'low', when it should have confidence 'low' and service confidence 'low' (overriden)
ftp will have confidence 'high' and service confidence 'high' (expected)
etc...

What we really want: when masscans happen, set global confidence to 'low' unless the service confidence is high and the service_name is not `tcpwrapped`.

So in the above example:
nfs (global_confidence -> high because service_confidence = high; bypassing the masscan-set global confidence)
tcpwrapped (global confidence -> low because it's tcpwrapped and the masscan-set global confidence is low; service confidence = low (downgraded from 'high')
ftp (global confidence -> high because service_confidence = high; bypassing the masscan-set global confidence)
rdp (global confidence -> low because masscan-set global confidence and service_confidence = 'low')

ocervell: Good catch — you were right, the shared `global_confidence` was still clobbered. Fixed in 9adc9a7: confidence is now computed **per port**, never carried across the loop.

Your example resolves exactly as you described:
- `nfs` (conf=10, not tcpwrapped) → **high** (bypasses the IDS 'low')
- `tcpwrapped` → **low**, and `service_confidence` **downgraded to low** (was leaking nmap's conf=8)
- `ftp` (conf=10) → **high**
- `rdp` (conf=1, non-tcpwrapped, *after* the confident ports) → **low** ← the clobber case, now correct

Rule on a mass-scan host: `low` unless `service_confidence == 'high'` and not `tcpwrapped`. Off an IDS host, ports stay `high`. Test `test_per_port_confidence_no_clobber` mirrors this exact nfs/tcpwrapped/ftp/rdp sequence and asserts the trailing rdp stays low.

PR #1401, opened 2026-09-17
fix(ai): never use platform LLM key with a custom api_base; force-inherit isolated to subagents

ocervell:
## What / why

Two verified security issues in the AI core task (authorized owner audit).

### V-AI-01 — platform LLM key exfiltration via `api_base`
A user-supplied `api_base` run-opt made the AI task send the **platform** LLM key
(`CONFIG.addons.ai.api_key`) to an attacker-chosen endpoint (`ai.py` `_init_options`
`api_base ... or CONFIG.addons.ai.api_base` / `api_key ... or CONFIG.addons.ai.api_key`
→ `call_llm`). That leaks the key and enables worker SSRF (`http://169.254.169.254/…`).

**Owner decision (not a blanket deny):** keep allowing `api_base`, but never hand the
platform key to a caller-**overridden** base. Credentials now resolve through
`resolve_llm_credentials()`:

- **default / unset / base == `CONFIG.addons.ai.api_base`** → platform key (unchanged).
- **base overridden (non-empty and `!= CONFIG.addons.ai.api_base`) + caller `api_key`** → the caller's own key.
- **base overridden without a caller `api_key`** → run **refused** (`ValueError`); the platform key is never sent.

Mirrors the existing subagent guard in `ai/actions.py` (pops a tool-supplied
`api_base`/`api_key` so it can't redirect the injected parent key).

### M1 — `isolated` (run_shell DinD sandbox) lowerable by a child
A child/subagent spawned by an AI task could set `isolated=false` and drop out of the
sandbox. `isolated` is now **force-inherited** from the parent in `_child_run_opts`
(`isolated: ctx.isolated`) and added to `_FORBIDDEN_CHILD_OPT_KEYS` so
`_sanitize_child_opts` strips any LLM/child-supplied value before the run_opts merge.
A child always runs at the parent's isolation level and can never weaken it. Core CLI
users may still set `--isolated`; only child propagation is locked (the secator-api
`DISALLOWED_RUN_OPTS` half of M1 is handled separately in that repo).

## Default-vs-override rule
`caller_api_base and caller_api_base != CONFIG.addons.ai.api_base` ⇒ overridden ⇒
require caller key. Empty caller base, or one equal to the configured base, is not an
override and keeps the platform key.

## Tests
- `test_ai_task_opts.py::TestResolveLlmCredentials` — default→platform key; base==platform→platform key; custom+own key→own key; custom without key→`ValueError` (no leak).
- `test_ai_isolation.py::TestChildInheritsIsolated` — child inherits parent `isolated` (true/false); LLM cannot set `isolated`; child cannot lower it through the real merge order.

Targeted subset green (20 passed). 6 pre-existing failures in
`test_ai_guardrails.py::TestEdgeCases` (redirect path classification) reproduce on
clean `origin/main` — unrelated to this change.

https://claude.ai/code/session_01Ls3QR8DxMkEeAX817KbHa3

## Summary by CodeRabbit

* **Security**
  * Child tasks now consistently inherit the parent task’s isolation setting, preventing sandbox overrides.
  * Platform LLM credentials are no longer sent to custom API endpoints. Custom endpoints require a caller-provided API key.
* **Bug Fixes**
  * Improved handling of LLM connection settings when using default or custom API endpoints.
* **Tests**
  * Added coverage for isolation inheritance and protection against credential leakage.


ocervell:
Superseded by #1407.

PR #1403, opened 2026-09-17
fix(core): quote/validate bbot presets and reject catastrophic mandate-scope regexes

ocervell:
Owner-authorized security audit follow-up. Two verified low/medium core-engine issues.

## CORE-4 — bbot `presets` argument injection (already fixed on main)
The audited vector — `presets` declared `shlex: False` so `_build_opt_str` never quotes it, letting a user-supplied value split into extra bbot argv tokens (argument injection into bbot, `shell=False`, not shell RCE) — **is already remediated on `main`** by PR #1309.

`secator/tasks/bbot.py` now carries an `on_cmd` construction-path hook that **allowlists** every preset token against `BBOT_PRESETS` and raises `ValueError` on an unknown token, so the injection is rejected before the command is ever built/run. This is the correct minimal fix: bbot legitimately needs multiple space-separated preset tokens as separate argv (that's why `shlex: False` + comma→space expansion exists), so simply `shlex.quote`-ing the whole value would break multi-preset usage — an allowlist keeps legitimate presets working while making arbitrary-flag injection impossible. It is covered by `tests/unit/test_task_opt_validation.py` (`test_bbot_rejects_flag_injection_via_presets`, `test_bbot_accepts_known_presets`). **No code change needed here.**

## M4 (core half) — catastrophic-backtracking mandate-scope regex
`secator/scope.py`'s compile-time ReDoS guard `_NESTED_QUANTIFIER` only caught a quantifier *inside* a quantified group (`(a+)+`). It **missed alternation-under-a-quantifier** like `(a|aa)+x` / `(foo|foobar)*` — overlapping alternatives under an unbounded quantifier, the other classic exponential shape. Such an entry passed the guard, reached `re.fullmatch()` and backtracked catastrophically on an attacker-controlled target (measured **16s on 40 chars**), stalling the single-process API event loop that runs the scope check.

**Fix:** add a companion `_ALTERNATION_QUANTIFIER` heuristic (`\([^()]*\|[^()]*\)[+*]`) and reject it in `_compile_entry` alongside the nested-quantifier case. Rejection is at **compile time** (fail-safe: `compiled=None` → the entry never matches and `fullmatch` is never called), so a pathological pattern can never trigger the backtrack. Best-effort and single-level like the existing guard; only unbounded `+`/`*` is rejected (a trailing `?` is bounded/safe), and a rejected ALLOW entry only narrows scope. A safe non-overlapping alternation like `([a-z0-9]|-)+` is a documented false positive — rewrite as a char class `[a-z0-9-]+`.

The secator-api half of M4 (running coverage off the event loop + input length cap) is handled by a separate secator-api PR and is **not** touched here.

## Test evidence
`tests/unit/test_scope.py::test_alternation_under_quantifier_is_rejected`:
- **without** the guard: hangs ~16s and fails the `<1.0s` assertion (catastrophic backtracking reproduced);
- **with** the guard: passes in <0.2s; normal regex (`app\.acme\.com`, `(dev|staging|prod)\.acme\.com`) and glob (`*.acme.com`) scope entries still compile and match.

`tests/unit/test_scope.py` + `tests/unit/test_task_opt_validation.py`: **30 passed**. (Unrelated pre-existing failures in `test_ai_guardrails.py::TestEdgeCases` are out of scope.)

https://claude.ai/code/session_01Ls3QR8DxMkEeAX817KbHa3

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection against regular expressions that could cause excessive processing or application slowdowns.
  * Unsafe patterns with overlapping alternatives and unbounded repetition are now rejected safely instead of being executed.
  * Standard scope patterns, including common alternatives and wildcard formats, continue to work as expected.


ocervell:
Superseded by #1407.

PR #1413, opened 2026-09-19
fix(ai): parse text/Hermes-style <tool_call> blocks when a model emits tool calls as content

ocervell:
## Problem

The AI agentic loop acts only on native structured `tool_calls`. Some models emit tool calls as **text** inside the message `content` in a Hermes/XML style, e.g.:

```
<tool_call>
<function=run_task>
<parameter=description> Run a nuclei scan
<parameter=task> nuclei
</tool_call>
```

When that happens litellm returns the text as `content` with `tool_calls` empty. The loop then (a) rendered the raw XML as a user-facing message and (b) hit its terminal branch (fires when `tool_calls` is empty), so it treated the turn as "no tool to run" and stopped/prompted. Net effect: the model's tool call was shown as raw text and the loop dead-ended.

## Fix

Add `parse_text_tool_calls()` in `secator/ai/utils.py`. When `tool_calls` is empty but `content` contains one or more `<tool_call>...</tool_call>` blocks, it parses them into the same litellm-shaped call objects the loop already consumes:

- function name from `<function=NAME>`, arguments from the `<parameter=KEY> VALUE` segments (tolerant of newlines/whitespace and missing closing tags);
- also accepts a JSON body (`{"name": ..., "arguments": {...}}`);
- parameter values are coerced from JSON when possible, else kept as plain text;
- handles multiple blocks; malformed/partial blocks are skipped, never raised.

The loop calls it right after `tool_calls = result.get("tool_calls", [])`, feeding recovered calls into the existing `_process_tool_calls` path exactly like native ones, and strips the consumed `<tool_call>` blocks from the displayed content so the raw XML is no longer shown. Native structured tool calls and plain content are untouched.

## Tests

New `TestParseTextToolCalls` in `tests/unit/test_ai_utils.py`: XML-style block parses + dispatches through `tool_call_to_action`, multiple blocks, JSON-style block, malformed block ignored, plain/empty content returned unchanged. Full AI unit suite (utils/loop/tools) green.

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing text-based tool calls emitted by compatible AI models.
  * Recovered tool calls are dispatched automatically, while recognized markup is removed from displayed response content.
  * Supports multiple calls and common XML- and JSON-style formats.

* **Bug Fixes**
  * Prevented raw tool-call markup from appearing in user-visible AI responses.

* **Tests**
  * Added coverage for valid, malformed, multiple, and plain-text responses.


ocervell:
Superseded by #1415 (consolidated).

PR #1414, opened 2026-09-19
fix(ai): guide guardrail-denied retries to in-scope host/IP and stop concurrent prompts from cancelling each other

ocervell:
Two small, independent AI-engine reliability fixes.

## (a) Guardrail retry guidance

When a network action is denied because only a *different form* of the host is in scope (the model used a raw IP but only the hostname is allowed, or vice versa), the model kept retrying the **same** denied target in a loop, burning turns.

Added concise guidance to the `<guardrails>` prompt (`secator/ai/prompts/constraints/guardrails.txt`):

> If a network action is denied because the target is out of scope, do NOT retry the identical target — it will be denied again. The allowed scope may list only one form of the host: if you used an IP, retry with the in-scope hostname; if you used a hostname, retry with the in-scope IP. Otherwise pick a different target that is in scope. Never repeat the same denied value.

## (b) Concurrent-prompt cancellation

**Report:** a user received a multi-choice follow-up prompt AND a permission/deny prompt at the same time; answering the deny prompt caused the follow-up prompt to stop awaiting an answer (orphaned).

**Root cause:** in the remote (DB-polling) backend, `_expire_stale_pending` is called whenever a new pending prompt is built (`build_pending_prompt`), to supersede a prior prompt left behind by a dead worker. But it expired **both** `follow_up` and `permission` pending docs indiscriminately. A permission and a follow-up prompt can legitimately be outstanding at the same time, so building/answering one flipped the other to `timed_out` — orphaning it. The existing test `test_build_pending_prompt_expires_prior_pending` had even codified this cross-type expiry.

**Fix (smallest safe correction):** scope the expiry to the incoming prompt's own type. Building a permission prompt now only supersedes a prior pending *permission*; building a follow-up only supersedes a prior *follow_up*. The intended same-type supersede (recovering a stale prompt from a mid-poll worker crash) is preserved; `_expire_stale_pending()` with no type still expires both (legacy callers).

**Residual (documented, out of scope here):** the answer-routing that resolves "the latest pending prompt" for a session lives in the API/persistence layer, not this engine. Poll correlation in this repo is already scoped per prompt via `extra_data.prompt_uuid`, so within the engine the two prompts now poll independently. A durable belt-and-suspenders follow-up (already FLAGged in the code) is a DB-layer TTL index on pending prompt docs.

## Tests

`tests/unit/test_ai_interactivity.py`: replaced the test asserting cross-type expiry; added tests that a permission build does not expire a pending follow_up, a follow_up build does not expire a pending permission, and that the legacy no-arg call still expires both. Full AI interactivity + prompts suites green. (6 unrelated `test_ai_guardrails.py::TestEdgeCases` redirect-detection failures are pre-existing in this environment — they need the `shfmt`/`safecmd` shell parser and fail identically on a pristine tree.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Pending permission and follow-up prompts no longer expire one another when both are active.
  - Prompt timeouts now apply only to the matching prompt type, preserving unrelated pending interactions.

- **AI Guidance**
  - When a network action is denied, the assistant now avoids retrying the identical target and prefers an allowed hostname, IP address, or alternative target when available.


ocervell:
Superseded by #1415 (consolidated).

PR #1416, opened 2026-09-20
feat(ai): proactive prefer-hostname scope prompt + optional hard-deny for out-of-scope targets

ocervell:
Follow-up to the surfaced-scope work: three AI-engine scope refinements.

- **Proactive prefer-hostname guidance.** The in-scope prompt line (`build_scope_section`) and the guardrails constraint now instruct the model up front to use an in-scope hostname rather than its resolved IP — scope is matched literally and a host's IP may be shared — instead of only reacting after a denial.

- **Optional hard-deny for out-of-scope network targets.** New `security.scope_hard_deny` flag (`SECATOR_SECURITY_SCOPE_HARD_DENY`, default `False`). Default behavior is unchanged: an out-of-scope network target still prompts for interactive approval (standalone CLI). When the flag is set, an out-of-scope target with a defined `in_scope` is denied outright with no prompt, so the model adapts to an in-scope target — for non-interactive deployments.

- **Structured out-of-scope deny.** An out-of-scope target denial (via the flag, or the existing mandate `out_of_scope` deny) now carries the offending target value and a machine-readable `out_of_scope` reason, so a UI/CLI can render a clear "Target X is not in the allowed scope" message and attach a retry action.

Tests: new `TestScopeHardDeny` (flag off asks; flag on denies with target+reason; in-scope still allowed; no-op without a scope) and a proactive-hostname prompt assertion. The touched AI suites (guardrails/prompts/interactivity/actions) pass, except the pre-existing `test_ai_guardrails.py::TestEdgeCases` redirect-parsing failures, which are unrelated (present on `main`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an optional hard-deny setting for out-of-scope AI network targets, configurable through security settings or an environment variable.
  * Denials now identify the specific target that falls outside the allowed scope.

* **Improvements**
  * AI guidance now prioritizes approved hostnames over resolved IP addresses unless the IP is explicitly allowed.
  * Out-of-scope targets continue to request approval by default when hard-deny is disabled.

PR #1422, opened 2026-09-21
fix(ai): clean errors for invented run_task/run_workflow names (no traceback leak)

ocervell:
## What
When the AI agent calls `run_task`/`run_workflow` with a name that isn't a valid runner, it now fails with a **clean, actionable message** instead of leaking a raw Python traceback into the model's context.

## Why (evidence from two canary AI runs, workspace `bb_21_arcbbc`)
The model invents runner names or calls a real **workflow** through `run_task`. Concretely it called:
- `run_task name="url_crawl"` and `run_task name="code_scan"` — both are real **workflows**, not tasks.

For a task, `TemplateLoader(input={'type':'task','name':name})` accepts *any* name; the miss only surfaces later inside `Task.build_celery_workflow -> get_task_class`, which raises `TaskNotFoundError` **during** `yield from runner`. So the existing construction-time `except TaskNotFoundError` (actions.py) never catches it, and the tool result carried a full stack trace:

```
{"status":"error","results":[...,{"message":"TaskNotFoundError: Task url_crawl not found. Aborting.","traceback":"Traceback (most recent call last):\n ... get_task_class ... raise TaskNotFoundError"}]}
```

Each miss burned an iteration and polluted context with a scary trace.

## Fix (root cause, up front in `_run_runner`)
- **task**: validate via `Task.get_task_class(name)` before construction. Unknown -> `Error` listing available tasks; if the name is actually a real workflow, tell the model to use `run_workflow`.
- **workflow**: an unknown name loads an empty template -> detect and `Error` listing available workflows.

No traceback, no dispatch, model can self-correct immediately.

## Tests
TDD in `tests/unit/test_ai_actions.py::TestRunRunnerNameValidation` reproducing the transcript failure (invented task, workflow-called-as-task cross-hint, unknown workflow, valid-name passthrough). Full `test_ai_actions.py`: 112 passed, 3 skipped.

## Scope
Not a permission/scope change — orthogonal to #1416 and secator-api #325. Based on `canary`.

https://claude.ai/code/session_016GVpccvLhu6mvX1NFPYoA1


ocervell:
Folded into #1427 (unified AI PR).

PR #1426, opened 2026-09-21
fix(ai): single-flight remote turns so redelivery can't replay concurrently

ocervell:
## Problem (canary, analysis `bb_21_arcbbc`)

A remote (web) `ai` Celery task **replayed the whole conversation** on redelivery: the identical initial prompt ran at **12:25 AND 12:43** on the same worker, and **3 concurrent run_ids replayed 13:18–13:24**. Root cause: an `acks_late` redelivery (or a late-answer respawn) of a still-in-flight turn ran **concurrently** with the original — same session, **no mutex** — so every completed turn re-executed and tokens were re-billed.

The existing completed-turn marker (`_turn_completed_marker`) only guards a turn that has **already finished**; a turn redelivered **while still running** has no marker yet, so nothing stopped a second delivery from piling on.

## Fix — per-session single-flight

- **`_run_remote_turn`** wraps the marker-check + resume/fresh run in **`_single_flight`**, so the lock is held across *both* the idempotency check and the run. A waiter that acquires the lock later re-observes the finished state (marker present) and **no-ops instead of replaying**.
- **`_single_flight`** — a **non-blocking** `redis-py` `Lock` keyed by **`session_id`** (token + Lua safe release). TTL = `broker_visibility_timeout`, so a crashed holder's lock expires **exactly when the broker redelivers** the task, letting the redelivery resume. A contended delivery no-ops; **different sessions still run in parallel**.
- **Fail-open**: no Redis backend (e.g. filesystem broker in dev) or any lock error → run unlocked, never wedge the run.
- **`_run_fresh_turn`** extracted from `yielder` — shared by the local path and the fresh remote path (now run while holding the lock).

**Legitimate resume-after-late-answer still continues**: a late answer respawns with a *new* `celery_id` → no matching completed marker → resumes from prior docs. Only concurrent duplicates are suppressed — the goal is CONTINUE, not REPLAY.

## Tests (`tests/unit/test_ai_session.py::TestSingleFlight`)

- `test_two_concurrent_invocations_only_one_runs` — two deliveries share one lock: the first runs the turn, the second finds it held and no-ops (never touches resume/fresh); after release a later delivery acquires and runs.
- `test_contended_delivery_short_circuits_before_resume`
- `test_fresh_remote_turn_runs_inside_lock`
- `test_fail_open_when_no_lock_backend`
- `test_make_session_lock_none_for_non_redis_and_local`
- `test_release_error_is_swallowed`

Completed-turn idempotency (items 1 & 2 of the brief) was already present on canary and remains covered by `TestTurnIdempotency`; this PR adds the missing single-flight layer (item 3).

`pytest tests/unit/test_ai_session.py` → **48 passed**. Full `tests/unit` → **0 AI failures** (63 pre-existing base failures in `test_config`/`test_cve`/`test_offline`, all network/config, unaffected by this ai.py-only change).

## Residual

None of this needs the RabbitMQ broker migration — the fix is contained to idempotency + single-flight. The RabbitMQ migration remains the separate, longer-term answer to `acks_late` redelivery in general; this PR makes the `ai` task safe against it in the interim.

https://claude.ai/code/session_016GVpccvLhu6mvX1NFPYoA1


ocervell:
Superseded by the consolidated AI-fixes PR #1427, which lands the resume/repeat root-cause fixes: RC1 (`_maybe_resume_remote` re-append guard `_prompt_already_tail`) and RC2 (bounded follow-up auto-extension via `_MAX_FOLLOWUP_EXTENSIONS`), plus end-to-end mock-LLM conversation tests. The single-flight/lock approach here isn't needed once the duplicate re-append and the unbounded answered-follow-up loop are fixed at the source. Closing in favor of #1427.

